PR #1, opened 2017-05-30
Improve infra

miguelinux:
Some warnings were showed when documentation was building.
These commits modify conf.py to not show those warning.

Also Makefile from the root path is improved, so now is pointing to the targets inside source directory.


rcaballeromx:
Thank you for your patience. I just gave Ianeta the training and can now merge these changes.

PR #2, opened 2017-06-01
cc: Add code contribution guidelines.

rcaballeromx:
The contribution guidelines define the code style and the contribution methods of the project.

### DIFF
--- a/source/clear-containers/clear-containers.rst
+++ b/source/clear-containers/clear-containers.rst
@@ -27,3 +27,4 @@ container infrastructure.
 
     getting-started
     architecture-overview
+    contribution
--- a/source/clear-containers/contribution.rst
+++ b/source/clear-containers/contribution.rst
@@ -1,0 +1,231 @@
+.. _cc-contribution:
+
+Contribute to the Intel® Clear Containers OCI runtime
+#####################################################
+
+The |CC| :abbr:`OCI (Open Container Initiative)` runtime is an open source
+project licensed under the `GPL v2 License`_ .
+
+Coding styles
+=============
+
+C code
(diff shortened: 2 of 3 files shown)


dlespiau:
We want to keep CONTRIBUTING.md inside the git repository. It has a special meaning on github.

rcaballeromx:
@dlespiau We will. It will have a link to this guideline.

sameo:
@rcaballeromx The link should go from the cc github repo to here, not the other way around. In other words, you want to improve https://github.com/clearcontainers/runtime/blob/master/CONTRIBUTING.md rather than adding a new one here.

rcaballeromx:
I know that a link to CONTRIBUTING.md is shown on issues and pull requests. That's why I do not propose we get rid of it. Since GitHub only provides a link, there is no need for the content to be there. A very short file pointing to the full guideline is sufficient and would be rendered perfectly by GitHub.
On the other hand, we need the content on the website to fulfill the following requirements:
* Comply with the requirements set by both external and internal UX experts regarding community and collaboration content needed for the website.
* Have the content appear on search engine results, which does not happen for the files in GitHub.
* Allow us to gather metrics on the content to gain very useful insights regarding number of new contributors, time they spend on the content, etc. Those metrics allow us to improve the content to provide better experiences and become useless if the bulk of the content is in the MD file.
Unfortunately, there is no way for me to just import the MD file into the website and having the content on both places creates maintenance problems.

rcaballeromx:
Based on the new documentation decisions, this content will be transformed to MD and added to the code repo instead. This PR is being closed without merging.

PR #4, opened 2017-06-02
network: Add guide for network bonding

bwarden:
Add a guide for network bonding, including systemd network
configuration.

Signed-off-by: Brett T. Warden <brett.t.warden@intel.com>

### DIFF
--- a/source/clear-linux/network/network-bonding.rst
+++ b/source/clear-linux/network/network-bonding.rst
@@ -1,0 +1,133 @@
+.. _network-bonding:
+
+Network Bonding
+###############
+
+Network bonding is a technique for combining multiple network interfaces into
+a single, logical interface, providing some combination of redundancy and
+bandwidth aggregation.
+
+|CLOSIA| includes the bonding_ and team_ drivers. Here, we'll provide an
+example of configuring systemd to use the ``bonding`` driver.
+
+All commands in this guide should be run as root.
+
+First, if it does not already exist, create the following directory:
+
+.. code-block:: console
+
+   # mkdir -p /etc/systemd/network
+
+In this directory, you will create the configuration files for the virtual
+device and the network settings for it and the underlying physical interfaces.
+
+Next, configure systemd to create a virtual network device, ``bond1``. The
+syntax for this file is defined in the systemd.netdev_ manpage.
+`This example`__ may be used verbatim, or tuned to your particular requirements.
+Note that ``802.3ad`` mode requires explicit support from your NICs and network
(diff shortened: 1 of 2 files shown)


mltullis:
Rodrigo is out WW 23, but I'm covering and will get this merged for the Monday doc build.

bwarden:
It's not urgent, but I would appreciate any reviews/feedback.

source/clear-linux/network/network-bonding.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Network bonding is a technique for combining multiple network interfaces into
+a single, logical interface, providing some combination of redundancy and
+bandwidth aggregation.
+
+|CLOSIA| includes the bonding_ and team_ drivers. Here, we'll provide an
+example of configuring systemd to use the ``bonding`` driver.
+
+First, if it does not already exist, as root, create the
+``/etc/systemd/network`` directory:
```
jodh-intel: Wouldn't it be simpler to say, "... create the directory below:" as that would avoid the duplication of that string (which makes maintenance of this doc easier)?

source/clear-linux/network/network-bonding.rst
```diff
@@ -0,0 +1,131 @@
...
+
+   [Network]
+   Bond=bond1
+
+   [Link]
+   MTUBytes=9000
+
+.. note::
+
+   I'm bonding all four ports of a quad-port NIC, with names in the range
```
jodh-intel: You have switched from the first person plural to singular here.

Related to this, is there a convention for these guides to not use the passive voice?

source/clear-linux/network/network-bonding.rst
```diff
@@ -0,0 +1,131 @@
...
+
+   Once again, I'm enabling jumbo frames of up to 9000 bytes. This option must be
+   enabled for all slave interfaces *and* the bonded interface in order to take
+   effect.
+
+To apply the new network configuration:
+
+.. code-block:: console
+
+   # systemctl restart systemd-networkd
```
jodh-intel: There is an implicit assumption here that this command should be run as `root` (due to the hash prompt). I think it would be clearer to state at the start of the guide that all commands that use this prompt should be run as `root` (unless there is a "meta-guide" somewhere that already explains this?)

mltullis:
@bwarden: I see that you've incorporated review comments from @jodh-intel. We'll merge now and then do some final minor changes (adding steps, making voice consistent) before the public build today.

PR #6, opened 2017-06-06
clear-linux: Update bootable-usb xzcat command option

bryteise:
Do not use -v with xzcat based on experience with it causing issues
since the verbose output sometimes gets piped to the dd command.


mltullis:
merged

PR #13, opened 2017-06-29
Correct update content validation commands

gtkramer:
To verify the Manifest.MoM, the signature of the Manifest.MoM should be compared with the Swupd certificate, not the Clear Linux certificate.  Also, adding a section for manually validating the Swupd certificate so that it matches the section for manually validating the Clear Linux certificate.  Making the distinction between the Clear Linux certificate and the Swupd certificate more clear.


tpepper:
lgtm

miguelinux:
lgtm

PR #24, opened 2017-08-15
Update the Docker release version.

rcaballeromx:
Updated the release version of Docker compatible with the runtime. Removed the
release number of the Clear Containers to avoid having to update it every week.

### DIFF
--- a/source/clear-containers/getting-started.rst
+++ b/source/clear-containers/getting-started.rst
@@ -3,14 +3,10 @@
 Clear Containers getting started guide
 ######################################
 
-
-Introduction
-============
-
-The Intel® Clear Containers v2.1.5 release is now available.
-This release enables executing existing Docker applications in the secure
-and fast Intel Clear Containers environment under Docker\* 1.12 and beyond
-via an :abbr:`Open Container Initiative (OCI)` compatible `runtime`.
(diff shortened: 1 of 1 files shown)


source/clear-containers/getting-started.rst
```diff
@@ -3,14 +3,10 @@
...
-Introduction
-============
-
-The Intel® Clear Containers v2.1.5 release is now available.
-This release enables executing existing Docker applications in the secure
-and fast Intel Clear Containers environment under Docker\* 1.12 and beyond
-via an :abbr:`Open Container Initiative (OCI)` compatible `runtime`.
+The Intel® Clear Containers enable executing existing Docker applications in
+the secure and fast Intel Clear Containers environment under Docker\*
+v17.05.0-ce and beyond via an :abbr:`Open Container Initiative (OCI)`
```
grahamwhaley: I'll just note the 'and beyond' is a little tricky to quantify, as we cannot predict what changes may come in Docker. But, it is as good as we can get here I think. Similarly with what would be 'previous' - we may still work with previous versions of docker, but is is hard to say without thorough testing of (all) older versions etc.

rcaballeromx: I agree, we can leave it like this for now and once we move to the new system the release version will update alongside the code.

PR #27, opened 2017-08-18
Restructure clear linux content into five sections

iphutch:
The five sections include: get started, concepts, guides, tutorials, and reference.
File names untouched except where files were moved to sit under more suitable section.
Toc-trees changed to ensure all content is accounted for and in the correct section.
Titles within files were changed where necessary for consistent style.
Get-started label changed with reference to old label also changed.


source/clear-linux/bundles/bundles.rst
```diff
@@ -1,5 +1,5 @@
-Bundles
-#######
+Reference
```
rcaballeromx: Several things here: This file lacks a starting label .. _reference:
Sometimes, changing the filenames and directories is unavoidable. The reference content about bundles is part of the Reference content section, therefore, changing the heading here is not enough. Here, I would create the reference directory, place a reference.rst file and include the bundles.rst file in that toctree. Otherwise, you create a disparity between the content and the structure.

PR #49, opened 2017-10-14
fix hyperv link

Xeleron:


tpleavitt:
Verified that the link is correct for the hyperv-lts bundle

PR #51, opened 2017-10-19
Remove Clear Containers content from clearlinux.org/documentation

rcaballeromx:
Clear Containers has created their own documentation in their own repos.
This content is therefore redundant and causes confusion.


iphutch:
I understand redirects are going to be put in place.

Can we ensure the content from clear-containers.rst is in the clearcontainers/runtime and we redirect there. I do not believe we have an equivalent doc in that repo.

rcaballeromx:
@iphutch The content on clear-containers.rst is not in the clearcontainers/runtime repo since it pertains to CC 1.0 and 2.x. Therefore, we must not ensure the content is maintained there.

rcaballeromx:
Differing the removal of the content until we have identified the redirects needed.

PR #52, opened 2017-10-19
Remove ciao content from clearlinux.org/documentation

rcaballeromx:
Ciao has created their own content in their repos. Therefore, this
content is redundant and causes confusion.


iphutch:
Two questions:
1. Are redirects going to be put in place for these URLS? 
2. What is the state of the CIAO docs that correspond/duplicate the content in the docs we deleting here?

rcaballeromx:
@iphutch We can place redirects to the equivalent content. As for 2, Ciao already has more up-to-date guides in their repo.

PR #59, opened 2017-11-11
Add note on requiring a wired connection

GrantSeltzer:
A wired connection is currently required to connect to the update server during installation. Having this be clear in documentation would be helpful.


tpleavitt:
This is already plainly stated on line 33 of this document.

GrantSeltzer:
Ah, I missed that. Sorry and thank you!

PR #64, opened 2017-11-30
On 2017-11-30:

bktan8:
--------------
1) Created the download-image.rst file in the Guides/Maintaince.  This is a hidden file and not visible under the Maintenance tree.  This talks about types of Clear Linux images, how to verify checksum, and how to uncompress for Linux, macOS, and Windows.  This is the base for other docs such as the bare-metal, live, bootable USBs (Linux, macOS, Windows), VMware docs...
2) Reworked the bootable-usb-linux.rst to re-use content from download-image.rst.
3) Reworked the bootable-usb-mac.rst to re-use content from download-image.rst.
4) Reworked the bootable-usb-windows.rst to re-use content from download-image.rst.
5) Reworked the live-image.rst to re-use content from bootable-usb-linux.rst and download-image.rst.
6) Reworked the bare-metal doc to re-use content from bootable-usb-linux.rst and download-image.rst.


source/clear-linux/guides/maintenance/download-image.rst
```diff
@@ -0,0 +1,192 @@
...
+
+The naming convention of a |CL| image filename is: 
+
+  .. code-block:: console
+   
+    clear-<version number>-<image type>.<compression type>
+
+The `<version number>` field specifies the version number.
+
+The `<image type>` field specifies the type of image and its corresponding file format.  The current list of images available are:
```
tpleavitt: This list of image types could become outdated quickly.  I wonder if we should make this a separate file that we can update independently

source/clear-linux/guides/maintenance/download-image.rst
```diff
@@ -0,0 +1,192 @@
...
+* `kvm.img`: Image intended for booting in a simple VM with start_qemu.sh.
+* `live.img`: A live boot image of Clear Linux.
+* `safety-enhanced-linux.img`: 
+* `vmware.vmdk`: Image for VMware.
+
+The `<compression type>` field specifies the compression type. Two types of 
+compressions are used, namely GNU zip (`.gz`) and XZ (`.xz`).
+
+The latest, validated, and official releases are found in the `image`_ repository.
+Older official releases and unofficial - and not fully validated - releases are
```
tpleavitt: There is no way to tell an official release from an unofficial release as far as I can see.  I would change this to "Older releases are found in the releases directory.", removing mention of official and unofficial altogether.

PR #68, opened 2017-12-15
Fix figures on Multi-boot tutorial.

rcaballeromx:
This document added screenshots without using the figure directive.
Thus, the captions did not have the correct CSS class.
I applied the figure directive and removed the substitutions.
The captions now have the appropriate CSS class in HTML.


rcaballeromx:
Yes the new website was going live and it could not wait.
Merging like this is the exception, never the rule though.

PR #69, opened 2017-12-15
Fix 2 apt-get commands missing "install" option.  

bktan8:
On 2017-Dec-15:
---------------
Fixed a couple of a apt-get commands that were missing the "install" option.


bktan8:
@rcaballeromx - Please review.  Also, this is my first fork.  let me know if there's any issue.  Thanks!

PR #76, opened 2018-01-07
Removed unnecessary brackets.

sirredbeard:
Removed unnecessary brackets from date.


rcaballeromx:
Absolutely correct. The brackets are unnecessary but are there for aesthetics.

sirredbeard:
It looked like it was supposed to be formatting, but it wasn't coming through on the website. 

screen shot 2018-01-11 at 1 12 39 pm

rcaballeromx:
I have merged your commit since the brackets are unnecessary and no one has complained about the change.

PR #78, opened 2018-01-15
update bundle name to add VNC viewer

JosePerezCarranza:
Update bundle to "desktop-apps-extras" instead of "desktop-apps" to install VNC viewer on Clear Linux environment.


rcaballeromx:
LGTM

PR #79, opened 2018-01-20
Increase virtual disk size

bktan8:
This doc shows how to increase the size of a pre-built Clear virtual disk image.
This doc can be referenced by all of the VIRTUAL docs (kvm, hyper-v, vmware, etc.).


source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
```diff
@@ -0,0 +1,103 @@
...
+
+As you can see, both methods show the pre-built Hyper-V image is about 8.5GB. 
+
+Increase the root partition size
+********************************
+
+To increase the image size of a pre-built image, follow these steps:
+
+#.	Shutdown your VM, if it's running.  
+#.	Use the appropriate hypervisor tool to increase the virtual disk size of
```
tpleavitt: Can you give an example of what type of tool you need to use for this?

bktan8: You would use the tool for your hypervisor.  I'm happy to add step-by-step for various VMs if you think this is a must-have.  I was trying to focus on the Clear aspect.

PR #81, opened 2018-01-29
Update web tutorial

jairglez:
The current version of the web-server-install tutorial has some inconsistencies with the latest Clear Linux default values for the Apache and PHP servers (tested on VERSION_ID=20490) that prevents to complete the steps as they are now, or may cause confusion to the user. This patch-set includes some updates in screenshots, paths, typo fixing and step changes to account for the differences and correct errors.


tpleavitt:
LGTM and I like the section of creating the tutorial in it's own directory to give more clarity to customization.

PR #90, opened 2018-02-26
Fix ssh username passing comment.

ahkok:
Using `root@` and then `-l clear` is ambiguous.


rcaballeromx:
@ahkok Thank you for fixing this.

PR #91, opened 2018-02-26
aws-web: remove mention of -l parameter from text

bwarden:
Changing explanatory text to match the given example command, as corrected in #90.

### DIFF
--- a/source/clear-linux/tutorials/aws-web/aws-web.rst
+++ b/source/clear-linux/tutorials/aws-web/aws-web.rst
@@ -198,9 +198,9 @@ other instances available, they are also listed but not selected.
       chmod 400 AWSClearTestKey.pem
 
 #. Copy the text highlighted in the :guilabel:`Example:` section that is
-   shown in :ref:`figure 11<fig-aws-web-11>`. Paste the copied text into your terminal and add
-   `-l clear` text to the end of the command string. Press the :kbd:`Enter`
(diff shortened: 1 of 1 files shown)


source/clear-linux/tutorials/aws-web/aws-web.rst
```diff
@@ -198,9 +198,9 @@ other instances available, they are also listed but not selected.
       chmod 400 AWSClearTestKey.pem
 
 #. Copy the text highlighted in the :guilabel:`Example:` section that is
-   shown in :ref:`figure 11<fig-aws-web-11>`. Paste the copied text into your terminal and add
-   `-l clear` text to the end of the command string. Press the :kbd:`Enter`
-   key to execute the command.
+   shown in :ref:`figure 11<fig-aws-web-11>`. Paste the copied text into your
+   terminal, change the text before the :command:`@` sign to the username
+   :command:`clear`, and press the :kbd:`Enter` key to execute the command.
```
rcaballeromx: Thank you for updating the text. However, the use of the \:command\: role is incorrect. That role is reserved for actual command line commands. The solution is simple though, you can remove \:command\: and leave the back-ticks \`\`. Thus the text will render as literal text.

bwarden: Thanks, updated.

PR #93, opened 2018-03-08
Copy edit for grammar, formatting, and text flow.

MCamp859:
Deleted prompt character. Replaced # with sudo.
Assigned bash role to code blocks with user entries.
Edited text for active voice and consistency between steps.

### DIFF
--- a/source/clear-linux/guides/network/custom-clear-container.rst
+++ b/source/clear-linux/guides/network/custom-clear-container.rst
@@ -1,10 +1,10 @@
 .. _custom-clear-container:
 
-Build a custom Clear Linux\* based Docker\* container image
-###########################################################
+Build a custom Clear Linux based Docker container image
+#######################################################
 
 The official base |CLOSIA| container image is published on Docker\* Hub and is
-updated on a regular basis. This section contains the steps to build a custom
+updated on a regular basis. This guide contains the steps to build a custom
 container image.
 
 Prerequisites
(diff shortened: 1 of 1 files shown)


MCamp859:
@nesiusra  Please assign an engineer from your team to do a technical review of this content, thanks.

source/clear-linux/guides/network/custom-clear-container.rst
```diff
@@ -1,43 +1,42 @@
...
-      $ sudo -s
+      sudo -s
 
 #. Verify Docker is installed and running.
 
-   .. code-block:: console
+   .. code-block:: bash
 
-      # docker info
+      sudo docker info
```
tpleavitt: Since the first command you do is to get root privileges using the `sudo -s` command, you don't need to use sudo for the rest of the commands, so you need to remove sudo from the commands listed

MCamp859: Thanks for the feedback. I will confirm the changes with @rcaballeromx, perhaps I did too much.

rcaballeromx: QA and I have reached an agreement. 
Please make sure the `sudo -s` prefaces all other commands. 
We will then add the dependency on the test cases so that it gets executed before the other commands, even under parallel execution.
Once the command is in place @MCamp859,  you can remove `sudo` from all other commands.

tpleavitt:
The latest changes look good!

PR #95, opened 2018-03-12
Edited for grammar, formatting, and text flow.

MCamp859:
Deleted prompt character. Replaced # with sudo.
Assigned bash role to code blocks with user entries.
Edited text for active voice and consistency between steps.

### DIFF
--- a/source/clear-linux/guides/network/ipxe-install.rst
+++ b/source/clear-linux/guides/network/ipxe-install.rst
@@ -21,7 +21,7 @@ the flow of information between a PXE server and a PXE client.
 .. figure:: ./figures/network-boot-flow.png
    :alt: PXE information flow
 
-   Figure 1: PXE information flow
+   Figure 1: PXE information flow.
 
 .. caution::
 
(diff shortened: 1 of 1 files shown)


MCamp859:
@nesiusra Please assign an engineer from your team to do a technical review of this content, thanks.

nesiusra:
@ahkok - could you please review this one?  Thx!

ahkok:
So this all looks OK. The use of `sudo` is definitely good, but I really do not like it at this scale, because the amount of commands that require it is so large it is becoming distracting.

I would rather see a general remark near the top that people need to have root permissions for all the commands in this guide, and therefore should prepend `sudo` or just use `sudo su -` once from the start.

Given the amount of commands and here documents, this is probably too complex and novice users will likely fail to properly execute the here document sections properly. We should consider creating shell scripts and shipping them in CLR or use indirect ways of describing what needs to be done, for instance we can say "create a file `/foo/bar` with the following content: <quote content>".

Other than that, this is fine, I don't see any technical issues with these changes.

rcaballeromx:
@ahkok Thank you very much for taking the time to review this content. I agree with you on the complexity and amount of commands. I concur that scripting some of the steps would be a great help and I will raise the issue with the documentation team. We are adding the sudo commands to enable document based testing. I am meeting with the QA team to see what alternatives we can use, including your recommendations. 
For the time being, I suggest merging @MCamp859's edits and we will revisit the use of sudo once we have reached a new agreement with QA. Is this acceptable @ahkok?

ahkok:
yes, I we can merge this.

PR #106, opened 2018-04-03
Added note to increase size of VM partition

user135711:
I added a note that might confuse new users when following the getting started sequence download, import vmware VM->enable user space. The linux error will not indicate that the hard drive isn't big enough. The link did not render in the preview for some reason so edit it to conform with your relative links convention and build before pulling.

### DIFF
--- a/source/clear-linux/guides/maintenance/enable-user-space.rst
+++ b/source/clear-linux/guides/maintenance/enable-user-space.rst
@@ -97,9 +97,10 @@ To be able to execute all applications with root privileges, we must add the
 
 Install a GUI to test sudo
 --------------------------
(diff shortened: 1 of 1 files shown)


Michael-Vincerra:
Thank you for your suggestion. We welcome this addition. We just need you to make a  minor change to the .rst syntax. Could you revise your file to use syntax as follows and then resubmit the PR?:

.. note::
   If you are following this sequence after just setting up the pre-configured VMware virtual machine from 
   the repo, you must :ref:` increase virtual disk size<increase-virtual-disk-size>` or the following step 
   will fail.

Michael-Vincerra:
Hello user135711. Thanks for your contribution. Will you please delete your branch? Thanks.

PR #119, opened 2018-04-24
Fixes Windows make.bat file for build process.

Michael-Vincerra:
Signed-off-by: Michael Vincerra <michaelx.vincerra@intel.com>

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -5,7 +5,7 @@ REM Command file for Sphinx documentation
 if "%SPHINXBUILD%" == "" (
 	set SPHINXBUILD=sphinx-build
 )
-set BUILDDIR=build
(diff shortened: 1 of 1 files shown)


rcaballeromx:
The commit message is misleading. You are not fixing the make.bat, since it is working without issue.
I suggest: update, modify, or similar.

Michael-Vincerra:
I guess my question is if 'build' folder should be set up outside 'source', shouldn't the current Makefile and .gitignore for clearlinux documentation be revised to reflect this?
 
https://github.com/clearlinux/clear-linux-documentation/blob/master/Makefile
 https://github.com/clearlinux/clear-linux-documentation/blob/master/.gitignore

I was using these, above, as a guide.

PR #120, opened 2018-04-24
Edited increase-virtual-disk-size guide for grammar, formatting, and …

MCamp859:
…text flow.

Removed # prompt and replaced with sudo in 4 instances.

### DIFF
--- a/source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
+++ b/source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
@@ -3,26 +3,26 @@
 Increase Clear Linux image virtual disk size
 ############################################
 
-Prebuilt |CLOSIA| images come in different sizes, ranging anywhere
-from 300MB to 20GB. This section shows you how to increase the size of your
-prebuilt |CL| image if you need more capacity.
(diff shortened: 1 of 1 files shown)


source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
```diff
@@ -39,88 +39,88 @@ Hyper-V* image version number 20450, you can follow these steps:
...
 		.. code-block:: console
 
 			Warning: Not all of the space available to /dev/sda appears to be
-			used, you can fix the GPT to use all of the space (an extra ... 
+			used, you can fix the GPT to use all of the space (an extra ...
 			blocks) or continue with the current setting?
 
 			Fix/Ignore?
 
-		Enter :command:`fix`.
```
mltullis: @MCamp859  --  Changes look good. The only question I have is if you intended to delete this line (Enter :command:`fix`). It might be an artifact of the diff, but I wanted to verify.

MCamp859: Thanks for checking -- the fix command was moved from line 96 to line 88.

PR #142, opened 2018-05-15
Removes ovs-dpdk per Rob Nesius; removes reference in parent toctree.

mvincerx:
Signed-off-by: Michael Vincerra <michaelx.vincerra@intel.com>


mvincerx:
@reaganlo FYI
See also PR 99
https://github.com/clearlinux/clear-linux-documentation/pull/99

PR #174, opened 2018-06-25
Update increase virtual disk size guide tutorial

puneetse:
Proposed formatting and clarifications to maintenance guide on increasing prebuilt Clear Linux disk size

Summary of changes in this PR:
* Reordered methods to find partition size to have the more likely one to be used first
* Added second level headers
* Added doctree table of contents with page anchors  
* Split single large figure into two

Please feel free to provide any feedback or suggested edits.


source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
```diff
@@ -1,126 +1,176 @@
...
+
+        sudo lsblk
+
+An example output of the :command:`lsblk` command: 
+
+    .. code-block:: console
+
+        NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
+        sda      8:0   0    8.5G  0 disk
+        ├─sdd1   8:1   0    512M  0 part
```
phmccarty: Shouldn't this line and the two below reference partitions from `sda` instead of `sdd`, since they are listed under `sda`? For example, I would expect this line to reference `sda1` instead.

mvincerx: @puneetse I have several comments and corrections to syntax. Can you check in with me in person?

mvincerx:
@puneetse I have several comments and corrections to syntax. Can you check in with me in person?

source/clear-linux/guides/maintenance/increase-virtual-disk-size.rst
puneetse: @phmccarty Hm, good catch. I'm not sure where that original output was captured from or why it was modified but you're right. I verified on another system to make sure sdd isn't referenced by default. 

@mvincerx done.

Issue #163, opened 2018-06-15
Consolidate documentation for image integrity and signature validation


phmccarty:
(Original report at https://github.com/clearlinux/distribution/issues/85)

Two pages already exist to document the related topics of verifying image checksum signatures and image integrity:
- [How to verify Clear Linux image integrity.](https://clearlinux.org/documentation/clear-linux/get-started/bare-metal-install#verify-the-integrity-of-the-clear-linux-image)
- [How to validate the signature of a Clear Linux image checksum file.](https://clearlinux.org/documentation/clear-linux/guides/maintenance/validate-signatures#image-content-validation) This section also covers image integrity verification, but a little differently.

It would be helpful for this documentation to be consolidated such that the steps are described in the same way.

intelkevinputnam:
Hi @phmccarty 

It looks like the version in the bare metal install is missing the verification of the certificate. Would you recommend simply including the section from the guide section in the bare metal? I've verified that the bare metal install works as advertised, but I haven't walked through the maintenance guide version.

phmccarty:
@intelkevinputnam Sure, I am okay with including the guide doc section on the bare metal page. The primary ask is to have the same information available (or linked to) on both pages, for consistency.

intelkevinputnam:
Added the following line and link to [Download, verify, and uncompress a Clear Linux image on Linux](https://clearlinux.org/documentation/clear-linux/guides/maintenance/download-verify-uncompress-linux), which is the source for the section in bare metal install: "For a more in-depth discussion of image verification including checking the certificate see [Image content validation](https://clearlinux.org/documentation/clear-linux/guides/maintenance/validate-signatures#image-content-validation)."

pohly:
Also note that there are technical issues with the openssl command (https://github.com/clearlinux/distribution/issues/85#issuecomment-395354903) and the current SHA512SUM file no longer works with the `sha512sum -c` command (https://github.com/clearlinux/distribution/issues/85#issuecomment-400601494).

The former might be worth mentioning in the documentation if it can't be fixed. The later probably should be fixed in the SHA512SUM files.

PR #178, opened 2018-06-26
Fix for issue #163 - lack of uniformity for validating image integrity

intelkevinputnam:
Most users will not need to check the certificate, but it is worth having a link to the certificate and the method for checking it should the user wish to do so. 

Details: Added a reference to validate-signatures.rst to download-verify-uncompress-linux.rst for uniformity (issue 163). Made a couple wording changes to validate-signatures.rst and added a label to help with navigation.

### DIFF
--- a/source/clear-linux/guides/maintenance/validate-signatures.rst
+++ b/source/clear-linux/guides/maintenance/validate-signatures.rst
@@ -1,7 +1,7 @@
 .. _validate-signatures:
 
-Validate signatures
-###################
+Validating signatures
+#####################
 
 |CLOSIA| offers a way to validate the content of an image or an update. All
 validation of content works by creating and signing a hash. A valid signature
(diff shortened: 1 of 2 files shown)


source/clear-linux/guides/maintenance/validate-signatures.rst
```diff
@@ -1,17 +1,19 @@
 .. _validate-signatures:
 
-Validate signatures
-###################
+Validating signatures
```
mvincerx: @intelkevinputnam. Minor revision request. Can you revert to original header here? 
We're trying to keep active language in headers too.  Otherwise, LGTM. Thanks for your work on this.

intelkevinputnam: Sure. Though this convention makes it difficult to refer to these sections elsewhere without having to change the wording.

mvincerx:
Thanks for your work on this.

PR #201, opened 2018-07-24
Adds target-anchors to cross reference with new method.

mvincerx:


intelkevinputnam:
Verified in local build. LGTM

PR #202, opened 2018-07-26
Appends ':' to the end of the target-label

mvincerx:
- where it appears in autospec-about
- where it is used in the "include"


intelkevinputnam:
LGTM

PR #206, opened 2018-07-31
Moved figure up a level to correct reference.

MCamp859:


mvincerx:
Thanks, all. LGTM.
https://github.com/MCamp859/clear-linux-documentation/blob/telem-fig-fix/source/clear-linux/concepts/telemetry-about.rst

PR #219, opened 2018-08-15
Lists hardware compatible with Clear Linux

iphutch:


source/clear-linux/reference/compatible-hardware.rst
```diff
@@ -0,0 +1,113 @@
+.. _compatible-hardware:
+
+Compatible Hardware
+###################
+
+This document describes hardware that has been tested and confirmed as
+compatible with |CLOSIA|. This list is not comprehensive and will continue to
+grow.
+
+.. list-table:: **Table 1. Clear Linux Compatible Hardware**
```
tpleavitt: A suggestion:  Can we group these by architecture?  Also, I don't see the latest Xeon SKL-x on here (Purley)

fenrus75:
can we do that in version two of this?

SKL-X is there but purley is the name of the reference motherboard not of
the cpu

iphutch:
OK folks, I believe this is done. Please review for v1 of the compatibility list.

PR #229, opened 2018-09-09
Add pundle descriptions and update bundles manually. 

mvincerx:


tpleavitt:
@mvincerx , I don't know Chris's GH acct and it isn't showing in the search.  I thought we had fully automated this listing?

PR #234, opened 2018-09-14
Applies Sphinx syntax and branding edits:

mvincerx:
- Revises headers to follow convention per collab guidelines
- Revises CL in accordance with new branding
- Removes `$` for CLI prompts and corrects code-blocks to bash
- Adds sentence in para 2, `download it here`, a previously unused link

### DIFF
--- a/source/clear-linux/tutorials/flatpak/flatpak.rst
+++ b/source/clear-linux/tutorials/flatpak/flatpak.rst
@@ -1,7 +1,7 @@
 .. _flatpak:
 
-Use Flatpak\* to install applications on Clear Linux\* OS
-#########################################################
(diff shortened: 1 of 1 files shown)


source/clear-linux/tutorials/flatpak/flatpak.rst
```diff
@@ -1,16 +1,16 @@
 .. _flatpak:
 
-Use Flatpak\* to install applications on |CLOSIA|
-#################################################
+Use Flatpak\* to install applications on Clear Linux\* OS
```
mltullis: I believe Kris set up and already merged new substitutions for the updated branding, so we should use that. This also applies to other instances in this file.

PR #246, opened 2018-09-26
updates to reflect telemetry client refactoring

bd-dean:
changes 'telemd' to telempostd & telemprobd.   Updates diagram to show new daemons.


source/clear-linux/concepts/telemetry-about.rst
```diff
@@ -3,7 +3,7 @@
 Telemetrics
 ###########
 
-One of the key features of |CLOSIA| is telemetry, which is used to
+One of the key features of |CL| is telemetry, which is used to
```
tpleavitt: I believe the first instance of this needs to be |CL-ATTR|

source/clear-linux/concepts/telemetry-about.rst
```diff
@@ -47,11 +47,13 @@ Architecture
 The telemetry client provides the front end of a complete telemetrics solution
 and includes the following components:
 
-*	**telemd**, a daemon that prepares the records to send to a telemetrics server or
-	spools the records on disk in case it cannot successfully deliver them.
-*	Probes that collect specific types of data from the operating system.
+*	**telemprobd**, a daemon that prepares the telemetry records and spools them on disk prior to delivery
+* **telempostd**, a daemon that takes the records, verifies and then sends to the telemetry server or spools on disk in case the server is unavailable.
```
tpleavitt: The records are already spooled aren't they?  telempostd just sends them along or deletes them if the get too old

source/clear-linux/guides/maintenance/telemetry-enable.rst
```diff
@@ -3,7 +3,7 @@
 Enable and disable telemetry in Clear Linux
 ###########################################
 
-|CLOSIA| includes a telemetry solution as part of the OS that records events
+|CL| includes a telemetry solution as part of the OS that records events
```
tpleavitt: |CL-ATTR| should be used here

source/clear-linux/guides/maintenance/telemetry-enable.rst
```diff
@@ -36,15 +36,15 @@ This adds the telemetrics-client to your system and you will automatically
...
 
 .. code-block:: bash
 
 	sudo telemctl start
 
-This enables and starts the :command:`telemd` daemon and your system will
+This enables and starts the :command:`telemprobd` and :command:`telempostd` daemons and your system will
 begin to send telemetry data to the server defined in the file
 :file:`/etc/telemetrics/telemetrics.conf`. If this file does not exist, the
 :command:`telemd` daemon will use the file
```
tpleavitt: should be telempostd

source/clear-linux/guides/maintenance/telemetry-enable.rst
```diff
@@ -3,7 +3,7 @@
 Enable and disable telemetry in Clear Linux
 ###########################################
 
-|CLOSIA| includes a telemetry solution as part of the OS that records events
+|CL| includes a telemetry solution as part of the OS that records events
 of interest and reports them back to the development team via the telemetrics
 daemon, **telemd**. This functionality is maintained in the
```
tpleavitt: Should reference the new daemons

source/clear-linux/concepts/telemetry-about.rst
bd-dean: It depends on the configuration setting -- if configured to store locally instead of send, then telempostd stores them on disk as specified.

tpleavitt: telemprobd stores them in the spool directory and they stay in that location and then telempostd with try to send them to the server if record_server_delivery_enabled is set to true and the server is alive.  Telempostd will also delete a record after it expires (with the config option record_expiry value) or keep the record if record_retention_enabled is set to true.

mvincerx:
@bd-dean As we discussed this morning, I can merge this PR after you and @tpleavitt come to an agreement on a proper description on `telemprobd`.  Then, I will open a new PR to correct attributions and make editorial changes. So, please don't let the latter issues block you.

PR #253, opened 2018-10-03
Fix title of security concept page

puneetse:
The |CL-ATTR| substitution includes "OS" causing it to currently appear as "Clear Linux* OS OS Security"


mvincerx:
I agree. I think the suggestion from @intelkevinputnam aligns title of `security.rst` with other items on Concepts landing page. 
@puneetse. Will wait for a new push.

mvincerx:
LGTM. Thank you, all.

PR #255, opened 2018-10-04
kubernetes: Add tutorial

joselamego:
This tutorial describes how to install, configure, and run the Kubernetes
system on Cler Linux OS using either the Kubernetes runtime interface
CRI-O and kata-runtime or Docker* and runc.

This PR supersedes both #245 and #248 

Signed-off-by: Jose Lamego <jose.a.lamego@intel.com>

### DIFF
--- a/source/clear-linux/tutorials/kubernetes.rst
+++ b/source/clear-linux/tutorials/kubernetes.rst
@@ -103,6 +103,12 @@ deployment and your security needs.
          "sudo systemctl mask `sed -r -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps`"
          with
          "sudo systemctl mask $(sed -n -e 's#^/dev/([0-9a-z]).#dev-\1.swap#p' /proc/swaps)"
+#. Create (or edit if it exists) the hosts file that kubernetes will read to locate
+   master's host:
+
+   .. code-block:: bash
+
+      echo "127.0.0.1 localhost `hostname`" >> /etc/hosts
 
 #. Configure the Kubernetes runtime interface, either:
 
--- a/source/clear-linux/tutorials/kubernetes.rst
+++ b/source/clear-linux/tutorials/kubernetes.rst
@@ -1,0 +1,318 @@
+  .. _kubernetes:
+
+Run Kubernetes\* on Clear Linux
+###############################
+
+This tutorial describes how to install, configure, and run the 
+`Kubernetes container orchestration system`_ on |CL-ATTR| using different
+container engines and runtimes.
+
+Kubernetes\* is an open source system for automating deployment, scaling, and
+management of containerized applications. It groups containers that make up an
+application into logical units for easy management and discovery.
+
+Runc and `Kata Containers`_' kata-runtime are runtimes that adhere to :abbr:`OCI
+(Open Container Initiative)` guidelines and work seamlessly with Kubernetes.
+`Kata Containers`_\* provide strong isolation for untrusted workloads or
+multi-tenant scenarios.  Runc and Kata Containers can be allocated on a
+per-pod basis so you can mix and match both on the same host to suit your needs.
+
+Prerequisites
+*************
+
+This tutorial assumes you have installed |CL| updated to the latest release
+on your host system. You can learn about the benefits for cloud orchestration
+of having an up-to-date system in the `Software Update documentation`_.
+For detailed instructions on installing |CL| on a bare metal system, follow
+the :ref:`bare metal installation tutorial<bare-metal-install>`.
+
+Before you install any new packages, update |CL| with the following command:
+
+.. code-block:: bash
+
+   sudo swupd update
+
+Install Kubernetes and CRI runtimes
+***********************************
+
+Kubernetes and a set of supported :abbr:`CRI (Container Runtime Interface)` 
+runtimes are included in the `cloud-native-basic`_ bundle. To install the 
+framework, enter the following command:
+
+.. code-block:: bash
+
+   sudo swupd bundle-add cloud-native-basic
+
+Configure Kubernetes
+********************
+
+This tutorial uses the basic default Kubernetes configuration for simplicity.
+You must define your Kubernetes configuration according to your specific
+deployment and your security needs.
+
+#. Enable IP forwarding to avoid "preflight check" errors:
+
+   Create (or edit if it exists) the file :file:`/etc/sysctl.d/60-k8s.conf`
+   and include the following line:
+
+   .. code-block:: bash
+
+      net.ipv4.ip_forward = 1
+
+   Apply the change:
+
+   .. code-block:: bash
+
+      sudo systemctl restart systemd-sysctl
+
+#. Enable kubelet service:
+
+   .. code-block:: bash
+
+      sudo systemctl enable kubelet.service
+
+#. Disable swap using one of the following methods, either:
+
+   a) Temporarily:
+
+      .. code-block:: bash
+
+         sudo swapoff -a
+
+      .. note::
+
+         Swap will be enabled at next reboot, causing failures in
+         your cluster.
+
+   or:
+
+   b) Permanently:
+
+      Mask the swap partition:
+
+      .. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


joselamego:
@mvincerx: Please check this PR branch.

mvincerx:
@joselamego Clean build so far.  👍 Please allow me about 30 minutes to review. I will reply then.

mvincerx:
@joselamego I confirm that this PR 255 is a clean build. I will email you other git help for the future. 
Thank you for incorporating others' feedback. 
I had one technical question on which I want your opinion. 
See:  https://github.com/clearlinux/clear-linux-documentation/pull/248#discussion_r222755148
After the above issue is clarified, I'm ready to merge this PR. 
Please note: After I merge this PR, I'll follow up with another PR to fix attributions and formatting. :)

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -0,0 +1,324 @@
...
+         "sudo systemctl mask `sed -r -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps`"
+         with
+         "sudo systemctl mask $(sed -n -e 's#^/dev/([0-9a-z]).#dev-\1.swap#p' /proc/swaps)"
+#. Create (or edit if it exists) the hosts file that kubernetes will read to locate
+   master's host:
+
+   .. code-block:: bash
+
+      echo "127.0.0.1 localhost `hostname`" >> /etc/hosts
+
```
mvincerx: @joselamego. Latest push looks good for the localhost `hosts`. 
 
I wonder if we should document that we had to manually add a file `environment` , as in `/etc/environment` in order for the proxy settings to be recognized. This could be added to *Troubleshooting* section.  

If we assume that a proxy.conf is created for crio.service.d/ we could add something like: 

`cp /etc/systemd/system/crio.service.d/proxy.conf /etc/` 
`mv proxy.conf environment`

I'd like your opinion on whether this should be added. We're faced with whether a user could successfully troubleshoot proxy issues if we don't add this information.

joselamego: @mvincerx: this has been a recurrent item for discussion. Some folks argue that proxy setting is out of the Clear Linux documentation scope, and I agree to some extend. However, I can foresee it will be a typical cause of failure, so my suggestion is to only add to the proxy configuration section the general way to verify that in the user's current terminal the proxies are appropriately set: to run `echo $http_proxy`, `echo $https_proxy` and `echo $no_proxy` and the warning that if the output is different than what is expected, the procedure will fail.
I've confirmed that the appropriate file to set the proxies in a CL system is `/etc/profile`. So, in your suggested approach, it should be `mv proxy.conf profile`
Thoughts?

mvincerx: @joselamego  Yes, I'm aware that whether to include proxy config information is frequently debated. 
I really like your suggestion of adding commands like `echo $http_proxy`, etc., and of explaining that " if the output is different than what is expected, the procedure will fail."  Will you please add this in the Troubleshooting section?
While it may seem elementary, if adding it supports our users, everyone wins.

mvincerx: I approve of this. @joselamego Thanks for your dedication to improving this tutorial to address potential errors that users may encounter.  I invite those who were reviewers on #248 to do one final review of #255.   

@mythi, @eadamsintel, @tpleavitt 
Please let us know if any further changes are needed by 16:30 PST today.
Note: I will follow up #255 with edits to clearly distinguish workflows: 
`k8s + Docker` vs. ` k8s +CRIO`

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -0,0 +1,337 @@
...
+   or:
+
+   b) Permanently:
+
+      Mask the swap partition:
+
+      .. code-block:: bash
+
+         sudo systemctl mask `sed -r -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps`
```
eadamsintel: This does not work for me.  I am copying from Chrome on Windows into Putty ssh but this should work.

$ sudo systemctl mask `sed -r -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps`
sed: -e expression #1, char 37: invalid reference \1 on `s' command's RHS
Too few arguments.

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -0,0 +1,337 @@
...
+      .. note::
+
+         On systems with limited resources, some performance degradation may
+         be observed while swap is disabled.
+         On some systems you may need to substitute the command 
+         "sudo systemctl mask `sed -r -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps`"
+         with
+         "sudo systemctl mask $(sed -n -e 's#^/dev/([0-9a-z]).#dev-\1.swap#p' /proc/swaps)"
+#. Create (or edit if it exists) the hosts file that kubernetes will read to locate
+   master's host:
```
eadamsintel: This is confusing to me because why can't one command work for everyone?

BTW the alternative command did not work for me either.
`
$ sudo systemctl mask $(sed -n -e 's#^/dev/([0-9a-z]).#dev-\1.swap#p' /proc/swaps)
sed: -e expression #1, char 33: invalid reference \1 on `s' command's RHS
Too few arguments.
`

mythi: What's in your ```/proc/swaps```?

mvincerx: Confirmed. :( I received the same error when running this command. "Too few arguments."

mythi: Hold on. Both of the commands are wrong. Also, only one command is enough. There's no "on some systems..."

```sudo systemctl mask $(sed -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps)```

mvincerx: @mythi. Thanks for your feedback.  Please let us know as soon as you find the correct command. 
Both commands fail when I try them too.

mythi: I have it posted above.

mythi: ```Too few arguments.``` means you already have swap disabled (from previous attempts?)

mvincerx: I talked with @eadamsintel. We discussed adding a note here: 

If `Too few arguments` appears as error message, determine if `swap` is already disabled.  
Run: `cat /process/swaps`

Do you agree this note should be added?

mythi: Or

```sudo systemctl mask $(sed -n -e 's#^/dev/\([0-9a-z]*\).*#dev-\1.swap#p' /proc/swaps) 2>/dev/null```

mvincerx: @joselamego I agree that only one command should be used. Remove the current `note` that begins with "On some systems...." Please replace existing command, after *Mask the swap partition* with @mythi 's revised command above. 

Please let me know if you think adding my note, in response to `Too few arguments`, is necessary.

joselamego: LGTM. @mvincerx: should I make the edit and push?

mvincerx: Yes, please!

mvincerx:
Thanks all for great collaboration and input on this PR. I'm approving and releasing now due to demand for tutorial. Note:  I will follow up with another PR on this tutorial for editorial revisions.

mvincerx:
Closes #248

mythi:
@mvincerx thanks! is this going to land on the website soon?

mvincerx:
Yes. I'll let you know. I'm working with a remote team to resolve some production build issues. Hold tight.

mvincerx:
https://clearlinux.org/documentation/clear-linux/tutorials/kubernetes

PR #268, opened 2018-10-11
fixed example URLs that were not supposed to be hyperlinks

intelkevinputnam:
bulk-provision.rst contains example URLs that were not meant to be links. Used `` to render them as inline code. Also made a small wording change to add clarity.


mvincerx:
LGTM! Thank you.

PR #275, opened 2018-10-22
Update compatibility check with updated product name.

krisdale:
- Updated example output from compatibility check script based on updated script
- Updated page title: simplified title, reduce overuse of product name


mvincerx:
LGTM

PR #279, opened 2018-10-25
fixed link to sphinx-doc.org

intelkevinputnam:
@mvincerx since you’re the last person to work on this, I figured I’d run it by you.

There is one broken link that impacts two files: rest.rst and cross.rst (since it includes content from rest.rst)

The link that’s broken is this one: http://www.sphinx-doc.org/en/stable/markup/index.html

I’d like to replace it with: http://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html


mvincerx:
LGTM. Thanks for finding it. I'd like to see the link checker when there's time.

PR #296, opened 2018-11-01
Kubernetes tutorial: Added details in Install pod network add-on section

MCamp859:
Reviewers: @mythi has provided content, which includes fix for [clearlinux/clr-bundles#81 issue](https://github.com/clearlinux/clr-bundles/issues/81)

### DIFF
--- a/source/clear-linux/tutorials/kubernetes.rst
+++ b/source/clear-linux/tutorials/kubernetes.rst
@@ -203,12 +203,35 @@ communicate.
 
 Check whether or not your add-on requires special flags when you initialize
 the master control plane. For example, if you choose the `flannel` add-on,
-then you must add :command:`--pod-network-cidr 10.244.0.0/16` to the `kubeadm
-init` command.
+then you must add the following to the `kubeadm init` command:
 
-Similarly, if you choose the `weave` add-on, then you must change the CNI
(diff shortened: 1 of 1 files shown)


source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -203,13 +203,36 @@ communicate.
...
+
+..  code-block:: bash
+
+    plugin_dir = "/usr/libexec/cni/"
+
+to:
+
+..  code-block:: bash
+
+    plugin_dir = "/opt/cni/bin"
```
eadamsintel: I also had to symbolically link my loopback directory for weave to work. There might be a different more appropriate way to do this though but this worked for me.

ln –s /usr/libexec/cni/loopback /opt/cni/bin/loopback

MCamp859: @eadamsintel  Please look at the last commit, lines 248-252 contain the ln command.

eadamsintel: Ok I missed that.  Still, I don't know that this is the best way to do this but it works for now. If Clear Linux gets updated where this step is no longer needed then we can remove at that point.

mvincerx:
Thanks for the dedicated collaboration, all. Great improvements for the network plug-ins and user options.

PR #299, opened 2018-11-01
Greengrass tutorial: Corrects Sphinx syntax errors for rendering.

mvincerx:


mvincerx:
@bd-dean. After your approval. I will do a manual build like we discussed. Thanks.

PR #312, opened 2018-11-08
initial version end to end add telemetry to your application

bd-dean:


source/clear-linux/tutorials/telemetry-e2e.rst
```diff
@@ -0,0 +1,294 @@
+.. _telemetry-e2e:
+
+
+|CL-ATTR| includes a telemetry and analytics solution (also known as telemetrics) as part of the OS, which records events of interest and reports them back to the development team using the telemetrics client daemons.
+End users can enable or disable the telemetry client component of |CL| and also redirect where records are sent if they wish to collect records for themselves by using their own telemetry backend server.
+
+This tutorial walks you through creating a telemetry enabled app on your local |CL| machine. The tutorial will show you how to send your own custom reports to any telemetry backend server.
```
tpleavitt: I would continue to use records instead of reports

source/clear-linux/tutorials/telemetry-e2e.rst
```diff
@@ -0,0 +1,294 @@
...
+
+.. note::
```
tpleavitt: Do we need to continue to state this?  Also, we don't control what a third party can put in their payloads, only the built-in probes in Clear adhere to the policy

source/clear-linux/tutorials/telemetry-e2e.rst
```diff
@@ -0,0 +1,294 @@
...
+
+   The telemetrics functionality adheres to `Intel privacy policies`_ regarding the collection and use of :abbr:`PII (Personally Identifiable Information)` and is Open Source. Specifically, no intentionally identifiable information about the user or system owner is collected.
+
+Prerequisites
+=============
+
+For this tutorial, start with a clean installation of |CL| on a new system
```
tpleavitt: This really isn't a prerequisite.  The only prereq is that they have the telemetrics bundle installed on their system and they can check to see if the bundle is installed using swupd bundle-list.

source/clear-linux/tutorials/telemetry-e2e.rst
```diff
@@ -0,0 +1,294 @@
...
+using the :ref:`bare-metal-install` getting started guide and:
+
+#. Choose the automatic install.
+#. Join the :guilabel:`Stability Enhancement Program` during the installation process to enable the telemetrics client components.
+
+You can also set up a telemetry backend server; you can use the same |CL| system we're working with for this tutorial, or setup on a separate system.  Follow the :ref:`telemetry-backend` guide to setup the server and redirect your telemetry records to it. Refer to the :ref:`telemetry-config` guide to configure the telemetry client and specify where to send the records.
```
tpleavitt: Since this is outward facing the assumption should be that someone doing this won't be able to get to the internal telemetry server.  We should probably suggest they either set up their own server on their system so they can use localhost or go into the conf file and turn off delivery to the server and turn on record retention to enable journaling:
record_server_delivery_enabled=false
record_retention_enabled=true

Thoughts?

source/clear-linux/tutorials/telemetry-e2e.rst
bd-dean: Agree! Good catch.

source/clear-linux/tutorials/telemetry-e2e.rst
bd-dean: I can clean this up -- make it clear that the user won't be able to access their records on the default server.  The next paragraph points people back to the telemetry config guide page to show how to set up for local retention.

source/clear-linux/tutorials/telemetry-e2e.rst
bd-dean: I can clean this up -- make it clear that the user won't be able to access their records on the default server.  The next paragraph points people back to the telemetry config guide page to show how to set up for local retention.

source/clear-linux/tutorials/telemetry-e2e.rst
bd-dean: I can pull this out -- especially since we do state it on the telemetry guide pages.

source/clear-linux/tutorials/telemetry-e2e.rst
mvincerx: FYI: I plan to work on this branch for syntax cleanup. So, please don't worry about making those edits.

mvincerx:
Hey @bd-dean.  I completed my commits for now. I'll wait until after you've made revisions per our meeting today where we identified 3 different use cases. Thanks for being open to talk this through.

source/clear-linux/tutorials/telemetry-e2e.rst
```diff
@@ -245,7 +245,7 @@ Once the installation is complete you can use your web browser and view the new
 You should see a web page similar to the one shown in figure 1:
 
 .. TODO fix links for figures
-.. figure:: telemetry-backend/figures/telemetry-backend-1.png
+.. figure:: ../guides/telemetrics/telemetry-backend-1.png
    :scale: 50 %
```
mvincerx: @bd-dean Minor detail. For all documents that have figures, we need a `figures` directory added in which the two `.png` files reside. Will you please add that figures dir, move the .png files inside it, and update the two links again?

bd-dean: done

PR #321, opened 2018-11-16
Update bare metal images

krisdale:
- Refresh images with updated product name
- Remove whitespace from rst file


mvincerx:
LGTM.

PR #322, opened 2018-11-17
Revises first two <tr>'s of table; includes <style> as test.

mvincerx:


mvincerx:
@mltullis FYI only.

PR #323, opened 2018-11-17
Removes extra <tr> at top; removes <style>, overwritten by parent css.

mvincerx:


mvincerx:
@mltullis FYI only.

PR #328, opened 2018-11-21
Remove unused CLOSIA substitution definition

krisdale:
A previous PR removed the last use of the CLOSIA substitution. This PR removes the definition of the substitution entirely.


mvincerx:
LGTM.

PR #330, opened 2018-11-29
Adding autoproxy content

intelkevinputnam:
Content is identical to earlier PR without the organizational changes that caused the site navigation to break down. Organizational changes will be added later.


mvincerx:
LGTM.

PR #332, opened 2018-12-03
Added python script to improve linkcheck automation

intelkevinputnam:
Changes to Makefile:
* Added line to linkcheck recipe to run "parse-link-check.py"
* Appended "-" to linkcheck launch so it continues after returning errors (broken links cause this)

File added:
* **parse-link-check.py** generates broken_links.html in source/_build/linkcheck with live hyperlinks for manually checking broken links - in many cases the URLs are correct but break as the request was not made from a browser - need to be manually checked to verify they aren't broken


mvincerx:
@intelkevinputnam. FYI: You didn't specify reviewer(s).

PR #336, opened 2018-12-05
Fix formatting on kernel modules doc

puneetse:
This PR fixes an indent formatting issue on the kernel-module document causing a text block to inadvertently appear in a code block.


mvincerx:
LGTM.

PR #334, opened 2018-12-04
Deep Learning Reference Stack benchmarking tutorial

bd-dean:

PR #337, opened 2018-12-05
Kubernetes tutorial revisions

mvincerx:
Revisions and additions:
- Removes section Configure and run Docker + runc
- Removes references to Docker or runc environment
- Adds required installation of containers-basic bundle

### DIFF
--- a/source/clear-linux/tutorials/kubernetes.rst
+++ b/source/clear-linux/tutorials/kubernetes.rst
@@ -4,23 +4,19 @@ Run Kubernetes\* on |CL-ATTR|
 #############################
 
 This tutorial describes how to install, configure, and run the
-`Kubernetes container orchestration system`_ on |CL-ATTR| using different
-container engines and runtimes.
+`Kubernetes container orchestration system`_ on |CL-ATTR| using CRI+O and
+kata-runtime.
 
 Kubernetes\* is an open source system for automating deployment, scaling, and
 management of containerized applications. It groups containers that make up
 an application into logical units for easy management and discovery.
 
-Runc and Kata Containers\* kata-runtime adhere to :abbr:`OCI (Open Container Initiative*)`
-guidelines and work seamlessly with Kubernetes. `Kata Containers`_ provide
-strong isolation for untrusted workloads or  multi-tenant scenarios. Runc and
-Kata Containers can be allocated on a  per-pod basis so you can mix and match
-both on the same host to suit your needs.
-
-This tutorial describes the following combinations:
-
-* Kubernetes with Docker and runc
-* Kubernetes with CRI-O and kata-runtime
+Kata Containers\* kata-runtime adheres to
+:abbr:`OCI (Open Container Initiative*)` guidelines and work seamlessly with
+Kubernetes. `Kata Containers`_ provide strong isolation for untrusted
+workloads or  multi-tenant scenarios. Kata Containers can be
+allocated on a per-pod basis so you can mix and match both on the same host
+to suit your needs.
 
 Prerequisites
 *************
@@ -40,13 +36,12 @@ Before you install any new packages, update |CL| with the following command:
 Install Kubernetes and CRI runtimes
 ***********************************
 
-Kubernetes and a set of supported :abbr:`CRI (Container Runtime Interface)`
-runtimes are included in the `cloud-native-basic`_ bundle. To install the
-framework, enter the following command:
+Kubernetes, a set of supported :abbr:`CRI (Container Runtime Interface)`
+runtimes, and networking plugins, are included in the `cloud-native-basic`_ and `containers-basic`_ bundles. To install this framework, enter the following command:
 
 .. code-block:: bash
 
-   sudo swupd bundle-add cloud-native-basic
+   sudo swupd bundle-add cloud-native-basic containers-basic
 
 Configure Kubernetes
 ********************
@@ -105,7 +100,7 @@ deployment and your security needs.
          On systems with limited resources, some performance degradation may
          be observed while swap is disabled.
 
-#. Switch to root to modify `hostname`:
+#. Switch to root to modify `hosts` file:
 
    .. code-block:: bash
 
@@ -127,49 +122,7 @@ deployment and your security needs.
 Configure and run Kubernetes
 ****************************
 
-This section describes how to configure and run Kubernetes with:
-
-* Docker and runc
-* CRI-O and kata-runtime
-
-Configure and run Docker + runc
-===============================
-
-#.  Enable the Docker service:
-
-    .. code-block:: bash
-
-       sudo systemctl enable docker.service
-
-#.  Create (or edit if it exists) the file
-    :file:`/etc/systemd/system/docker.service.d/51-runtime.conf` and include the following lines:
-
-    .. code-block:: bash
-
-       [Service]
-       Environment="DOCKER_DEFAULT_RUNTIME=--default-runtime runc"
-
-#.  Create (or edit if it exists) the file :file:`/etc/systemd/system/kubelet.service.d/10-kubeadm.conf` and include the following lines:
-
-    .. code-block:: bash
-
-       [Service]
-       Environment="KUBELET_EXTRA_ARGS="
-
-#.  Enter the commands:
-
-    .. code-block:: bash
-
-       sudo systemctl daemon-reload
-       sudo systemctl restart docker
-       sudo systemctl restart kubelet
-
-#.  Initialize the master control plane with the command:
-
-    .. code-block:: bash
-
-       sudo kubeadm init --ignore-preflight-errors=SystemVerification
-
+This section describes how to configure and run Kubernetes with CRI-O and kata-runtime.
 
 Configure and run CRI-O + kata-runtime
 ======================================
@@ -211,8 +164,7 @@ If you choose the `flannel` add-on, then you must add the following to the
 
     --pod-network-cidr 10.244.0.0/16
 
-If you are using CRI-O and `flannel` and you want to use Kata Containers, edit
-the :file:`/etc/crio/crio.conf` file to add:
+If you are using CRI-O and `flannel` and you want to use Kata Containers, edit the :file:`/etc/crio/crio.conf` file to add:
 
 ..  code-block:: bash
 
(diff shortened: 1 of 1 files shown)


mvincerx:
@eadamsintel, @mythi, Hold on review of this document for now. I will request shortly. This PR is dependent on PR 334. 

@DnPlas mentions that " [kubernetes tutorial] does not show how to configure runc + CRIO for a kubernetes deployment.".. This should be added. Let's discuss this during first review.

eadamsintel:
Just glancing at this I don't think there are any real changes you need to do to get this to work with runc. The steps are the same. You specify in your yaml file for your pod which runtime to use and the default is runc as described. The tutorial does not actually explain how to run a kata pod and should especially with the new runtime class.

mvincerx:
Hello all. I'm requesting an expedited review of this tutorial.  Could you please review and comment/approve this PR by EOB today 17:00 PST?

The reason is that the Deep Learning as a Service (DLaaS) tutorial depends on the K8S tutorial in this PR.  See also: https://github.com/clearlinux/clear-linux-documentation/pull/334.  Management review of the DLaaS documentation is 09:00am PST Friday. 

Thanks in advance.

mvincerx:
@eadamsintel. Let's discuss running pods with K8S in a separate PR.

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -221,17 +173,11 @@ the :file:`/etc/crio/crio.conf` file to add:
...
-because it installs itself in the :file:`/opt/cni/bin` directory.
-
-If you are using Docker and `Weave Net`, edit the :file:`kubeadm.conf` file to
-add:
-
-..  code-block:: bash
-
-    Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
+If you choose the `Weave Net` add-on, then you must make the following
+changes because it installs itself in the :file:`/opt/cni/bin` directory.
```
eadamsintel: I am pretty sure the default crio.conf now has  plugin_dir = "/opt/cni/bin" which means I don't think it would work for either flannel or weave plugins without some further configuration.  Manohar's script at https://github.com/clearlinux/cloud-native-setup/blob/master/clr-k8s-examples/setup_system.sh provides clues on what users need to do.

I *believe* this command will solve the issue for both Weave and flannel and calico and other network overlays.    Please have Mikko confirm this is the correct final approach as I don't know what other changes to clear bundles might come.

sudo ln -s /usr/libexec/cni /opt/cni/bin

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -257,8 +203,8 @@ Use your cluster
```
eadamsintel: I don't this this is right anymore. I installed Clear Linux on 12/5 using that version and the default plugin_dir was "/opt/cni/bin"

eadamsintel: Also, I think the command from Manohar would work for every installation but I have not tested.  I don't think linking just the loopback would be adequate or rather if you use the command below I believe it will allow plugin_dir to be /opt/cni/bin and flannel and weave would just work.  I think this entire section could be collaped just to the below command.  You might want to put some text along the lines of ... Clear Linux installs CNI plugins that are part of the cloud-native-basic bundle to /usr/libexec/cni because that is a directory that Clear Linux's installer swupd can manage to ensure users can use swupd's verify feature to repair a system back to a known good state. Some CNI plugins like Weave auto install themselves to /opt/cni/bin.  To make sure that all of the common CNI plugins work smoothly on Clear Linux symbolically link the two directories."  something like that

sudo ln -s /usr/libexec/cni /opt/cni/bin

I have not tested this command on anything other than flannel.

eadamsintel: OK I need Mikko or someone else to verify. That command did not work for me when trying to create a weave network and the existing command "sudo ln -s /usr/libexec/cni/loopback /opt/cni/bin/loopback" also did not work.

kubeadm v1.12.3   crio version v1.12.3

From journalctl
Error adding network: failed to find plugin "loopback" in path [/opt/cni/bin/ /opt/loopback/bin]
Error while adding to cni lo network: failed to find plugin "loopback" in path [/opt/cni/bin/ /opt/loopback/bin]
Error deleting network: failed to find plugin "portmap" in path [/opt/cni/bin/ /opt/weave-net/bin]

mvincerx: @mythi, @eadamsintel, please see latest commit and provide comments. I replaced/collapsed  the above command, and I took inspiration from @eadamsintel to include a `note` below the newly added command.

source/clear-linux/tutorials/kubernetes.rst
mvincerx: Alternatively, like @eadamsintel mentioned, we could have users run Manohar's Bash script. Thoughts?

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -40,13 +36,12 @@ Before you install any new packages, update |CL| with the following command:
 Install Kubernetes and CRI runtimes
 ***********************************
 
-Kubernetes and a set of supported :abbr:`CRI (Container Runtime Interface)`
-runtimes are included in the `cloud-native-basic`_ bundle. To install the
-framework, enter the following command:
+Kubernetes, a set of supported :abbr:`CRI (Container Runtime Interface)`
+runtimes, and networking plugins, are included in the `cloud-native-basic`_ and `containers-basic`_ bundles. To install this framework, enter the following command:
```
ganeshmaharaj: do we need `containers-basic` to run crio and kata? the recent change in bundles seems to have included all the bits into `cloud-native-basic` itself.

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -211,54 +164,61 @@ If you choose the `flannel` add-on, then you must add the following to the
 
     --pod-network-cidr 10.244.0.0/16
 
-If you are using CRI-O and `flannel` and you want to use Kata Containers, edit
-the :file:`/etc/crio/crio.conf` file to add:
+If you are using CRI-O and `flannel` and you want to use Kata Containers, edit the :file:`/etc/crio/crio.conf` file to add:
```
ganeshmaharaj: i suspect this will apply outside of flannel too, but i am guessing flannel is the one you have tested and hence the statement?

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -211,54 +164,61 @@ If you choose the `flannel` add-on, then you must add the following to the
...
 
 ..  code-block:: bash
 
     [crio.runtime]
     manage_network_ns_lifecycle = true
 
-**Notes about Weave Net add-on**
+.. TODO: Should Weave and Calico be mentioned below, or is Flannel enought?
```
ganeshmaharaj: might be  good to tell the user that they can use any networkk plugin they choose, just that we have not tested with all of them and if they do find issues they can report a bug in our github? i am trying to avoid giving them the impression that kata will only work with flannel.

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -4,23 +4,19 @@ Run Kubernetes\* on |CL-ATTR|
 #############################
 
 This tutorial describes how to install, configure, and run the
-`Kubernetes container orchestration system`_ on |CL-ATTR| using different
-container engines and runtimes.
+`Kubernetes container orchestration system`_ on |CL-ATTR| using CRI+O and
```
ganeshmaharaj: i am wonder if it would be nice to say that things might also work with runc and docker and nothing here means that it will not. just that this is a tutorial for kata. not all users of clear will need kata for their k8s cluster.

source/clear-linux/tutorials/kubernetes.rst
mcastelino: containers basic is needed only for weave. Canal and Calico work with just cloud-native-basic

source/clear-linux/tutorials/kubernetes.rst
```diff
@@ -171,31 +171,45 @@ If you are using CRI-O and `flannel` and you want to use Kata Containers, edit t
...
+..         plugin_dir = "/opt/cni/bin"
 
-#.  Add the `loopback` CNI plugin to the plugin path with the command:
+.. #.  Add the `loopback` CNI plugin to the plugin path with the command:
 
-    ..  code-block:: bash
+..     ..  code-block:: bash
 
-        sudo ln -s /usr/libexec/cni/loopback /opt/cni/bin/loopback
+..         sudo ln -s /usr/libexec/cni/loopback /opt/cni/bin/loopback
```
mcastelino: This is the reason we need container-basic as that provides the CNI plugins today (I think).
Longer term we should just add the CNI plugins to container-native-basic.

@bryteise can we add CNI plugins to container-native-basic

It is preferrable not to have container-basic as docker adds its own iptable rules.
No having docker for k8s, would make the system cleaner

bryteise: I've submitted a patch for review adding CNI to cloud-native-basic.

source/clear-linux/tutorials/kubernetes.rst
mvincerx: @ganeshmaharaj , @mcastelino. I've removed `containers-basic` from this tutorial. For now, this tutorial only has to show it works with `flannel`.  In case you're not aware, I'm trying to expedite this PR because the Deep Learning tutorial, PR 334, depends on it, and on its use of `flannel`. 

I will expect to see changes to CL bundle for `Weave` next week--so I won't include it in this PR. I can incorporate bundle changes next week per @bryteise.

ganeshmaharaj: +1. we can always make changes later. this round need not cover all the things.

source/clear-linux/tutorials/kubernetes.rst
mythi: What was wrong with my patch sent a week ago? ;)

source/clear-linux/tutorials/kubernetes.rst
mvincerx: I will add a note:

You may choose to user other network plugins; however, they are not validated in the scope of this document.

source/clear-linux/tutorials/kubernetes.rst
mvincerx: Yes, we're restricting this tutorial's plugin to `flannel` only. @eadamsintel and I plan to do a mapping session for K8S tutorials during WW 51--during which we'll discuss which CNIs to incorporate and where in docs. Perhaps you'd like to call into that meeting?

ganeshmaharaj: would love to be there. Thanks @mvincerx

mvincerx:
Please note:  This PR will be merged 12/08/18 Saturday 15:00 PST. If you have any final edits on this PR, make them before this time. 

An upcoming launch of https://clearlinux.org/documentation/clear-linux/tutorials/dlrs *depends on the release and the accuracy of this kubernetes tutorial*; it must prove successful with `flannel` plugin ONLY. 

Use with other plugins can be addressed in later PRs.

PR #341, opened 2018-12-11
updating docker hub links

bd-dean:


source/clear-linux/tutorials/dlrs.rst
```diff
@@ -199,8 +199,8 @@ Run a TFJob
 
    .. code-block:: bash
 
-      ks generate dlrs-resnet50 dlaasresnet50 --name=dlrsresnet50
-      ks generate dlrs-alexnet dlaasalexnet --name=dlrsalexnet
+      ks generate dlrs-resnet50 dlrsresnet50 --name=dlrsresnet50
+      ks generate dlrs-alexnet dlrsalexnet --name=dlrsalexnet
```
DnPlas: If we do not change this, what will happen is that the worker pods won't start because on the next set of instructions we ask the user to run "ks apply default -c <component name>", where <component name> should be the same in both commands. All in all, it won't start.

bd-dean:
Ready to merge but need to wait for confirmation on new docker.hub build. Will confirm shortly, should be ready to merge later this afternoon.

Issue #314, opened 2018-11-09
Add Alternative to Multiboot to Chain Grub


mmccool:
**Describe the error/improvement to an existing document or image**
The current multiboot install instructions suggest disabling grub and modifying the menus in clr-boot-manager to boot other distributions using grub like Ubuntu.  However it's simpler to simply add a single menu item to let the clr-boot-manager to invoke grub if selected.   This also lets the target OS update its kernels, options etc. without having to change the configuration files for clr-boot-manager.   It also makes the configuration files for clr-boot-manager simpler since you don't have to copy over all the options, etc.   This does not even significantly slow down autonomous boot if you give both of them short timeouts (I use 5s for both).

For example, in my system, where I dual-boot an ubuntu installation in this way, I added only the following in `/boot/loader/entries/ubuntu.conf`:

    title Ubuntu
    efi /EFI/ubuntu/grubx64.efi

And now I never have to touch this again.

**Screenshots**
If applicable, add screenshots to help explain the error or unexpected behavior.

**Environment (please complete the following):**
NAME="Clear Linux OS"
VERSION_ID=26170

**Additional context**
There is a related issue with swupd that clears the timeout on the boot manager when you do an update.   There is an easy fix but I thought I would mention it here since it will affect people who are doing multiboot so maybe the workaround should also be documented in the same place:
https://github.com/clearlinux/swupd-client/issues/683

linjiach:
Thanks for the suggestion! It's indeed much simpler. We will update tutorial.

mvincerx:
Closed per GH-820

PR #343, opened 2018-12-12
Issue #314 Add Alternative to Multiboot to Chain Grub

linjiach:
Tested with Clear + Ubuntu 18.04 LST

https://github.com/clearlinux/clear-linux-documentation/issues/314

Signed-off-by: Linjia Chang

### DIFF
--- a/source/clear-linux/tutorials/multi-boot/multi-boot-mint.rst
+++ b/source/clear-linux/tutorials/multi-boot/multi-boot-mint.rst
@@ -61,47 +61,26 @@ tutorial.
 
    #. Log in.
 
-   #. Locate the Mint :file:`grub.cfg` file in the :file:`/boot/grub/`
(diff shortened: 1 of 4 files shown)


mvincerx:
FYI: I will test the multi-boot Ubuntu 18.04 use case .

intelkevinputnam:
I'll test the multi-boot Mint usecase.

intelkevinputnam:
If the intent is to enable side-by-side booting of both Mint and Clear Linux from the GRUB boot menu then the tutorial is insufficient. As written, it is not possible to boot into Clear Linux, as the Clear Linux boot option isn't added to the GRUB  boot list. Unless, I missed something, that is.

mvincerx:
Per Clear Linux docs meeting, the multi-boot document will be refactored to include only scenarios with Ubuntu, Fedora, and Windows 10; therefore, this PR will be closed (some content here to be used in new PR). Reference CLEAR-3987.

PR #345, opened 2018-12-14
Update links to FaaS source to point to version 1.0

bd-dean:


phmccarty:
@bd-dean Can you add some details to the commit about why the FaaS repo being is locked to v1.0? Is it because the latest version breaks compatibility?

bd-dean:
> @bd-dean Can you add some details to the commit about why the FaaS repo being is locked to v1.0? Is it because the latest version breaks compatibility?

Exactly that reason.  From the request: "... v2.0 of FaaS, which is NOT compatible with OpenVINO R3 released with Clear Linux’s computer-vision-basic bundle."

phmccarty:
@bd-dean Okay, thank you! Looks good to me.

mvincerx:
Built, tested, passed. Thank you both for staying on top of this document. Merging.

PR #351, opened 2018-12-17
Add scripts/_python dir to prepare to load Python scripts.

mvincerx:
New `scripts/_python` directory is intended as location for Python scripts,
which can be called from the parent-level source/Makefile. 


mvincerx:
@mltullis Thank you!

PR #366, opened 2019-01-16
change to getstarted to improve compatibility with sitenav

intelkevinputnam:
Right now the sitenav shows nothing beneath the pre-install and everything else appears below install. The sitenav needs toctrees to accurately represent the structure. This change will make sure the appearance of the get-started page matches the structure seen on the page with the exception of system requirements, which is a ref.

### DIFF
--- a/source/clear-linux/get-started/get-started.rst
+++ b/source/clear-linux/get-started/get-started.rst
@@ -11,31 +11,23 @@ Pre-install
 ***********
 
 * :ref:`system-requirements` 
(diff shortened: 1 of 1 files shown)


mvincerx:
@intelkevinputnam. Timing issue. I'll do a full review of this on Thurs.  I made significant structural changes to the `bootable-usb` dir in GH 358.  I'd like that PR merged before this PR is. Will explain more on Thurs.

PR #374, opened 2019-01-25
fixed minor bug in parse-link-check.py

intelkevinputnam:
This fixes a failure to convert int to string for output when an unexpected bad link is detected.


mvincerx:
Ready to merge? Recasting as a str seems like a simple fix.

intelkevinputnam:
Yep. Thx.

Issue #249, opened 2018-10-01
Document custom kernel modifications


tpleavitt:
From distribution created by [shrmrf](https://github.com/shrmrf) : clearlinux/distribution#215  

**Is your feature request related to a problem? Please describe.**
Documentation which describes (tutorial style perhaps) modification of the Linux kernel, adding and signing kernel modules is missing.

**Describe the solution you'd like**
- Tutorial on how with best-practices for Linux Kernel modification
- Kernel module signing documentation etc.
- Compiling a kernel module to run against Clear Linux

**Describe alternatives you've considered**
NA

**Additional context**
NA

shrmrf:
@tpleavitt cool, thanks!!

mrkz:
@tpleavitt @miguelinux would be worth to add to this documentation also how to add custom cmdline kernel parameters.

miguelinux:
We use those targets ;-)

`menuconfig`, `oldconfig` , `olddefconfig`

tpleavitt:
I asked one of our engineers about this and he replied as follows:

Our kernel packages include a configuration file that is fed into the build (https://github.com/clearlinux-pkgs/linux/blob/master/config) and when the kernel is installed in the system, the config is also stored in /usr/lib/kernel. If you're building the kernel yourself, you can certainly use `menuconfig` to modify the config file.

As part of the build, we run `make olddefconfig` to ensure that the ingested config is complete and correct. The copy that is stored in the package repo, which is input to this step, is typically generated through an automated process that similarly ends with `olddefconfig`, but this file is also subject to manual verification before we commit it.

shrmrf:
Thanks @tpleavitt + @miguelinux  can you elaborate a bit more? 

Maybe in tutorial style format? Where do we start (which repo?) and how can we use `make menuconfig` with it.. etc. etc.??

tpleavitt:
We added this document a few weeks ago and will continue to review other documentation that can be added to the website specific to kernel modifications.
[Add kernel modules](https://clearlinux.org/documentation/clear-linux/guides/maintenance/kernel-modules)

mvincerx:
@tpleavitt. Checking in.  Do you want to leave this issue open, or did we satisfy requirements?

tpleavitt:
@shrmrf , can you take a look at [PR#383](https://github.com/clearlinux/clear-linux-documentation/pull/383) mentioned above to see if this satisfies your request?

shrmrf:
[...]

PR #381, opened 2019-01-30
create linux-kernel.rst

rebeccasf:
First draft on kernel development guide with steps to setup tooling framework and show steps to modify linux kernel configuration for development use.

PR #383, opened 2019-02-01
Add document on kernel development for Clear Linux

puneetse:
This PR is adds a document describing how to create a custom kernel using the Clear Linux packaging tools. 

This PR is related to and aims to replace: https://github.com/clearlinux/clear-linux-documentation/pull/381

For reference, these are similar topics in the CentOS and Ubuntu documentation:
https://wiki.centos.org/HowTos/Custom_Kernel
https://wiki.ubuntu.com/Kernel/BuildYourOwnKernel
https://wiki.gentoo.org/wiki/Kernel/Configuration
https://wiki.gentoo.org/wiki/Kernel#Kernel_command-line_parameters


source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,335 @@
...
+   :ref:`autospec`, the kernel is not. This means control files provided 
+   by autospec are not available  and changes must be made directly.
+
+#. Open the Linux kernel package RPM SPEC file in an editor.
+
+   .. code-block:: bash
+
+      $EDITOR linux.spec
+
+#. Modify the Version, Release, and Source0 URL entries at the top of the file.
```
tpleavitt: What should I change these 3 fields to?  I would give an example of before and after.  Do I need to stay within a specific convention for version and release?  For source0, since I'm making a local change would I point to my local source?

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,335 @@
...
+       License:        GPL-2.0
+       Summary:        The Linux kernel
+       Url:            http://www.kernel.org/
+       Group:          kernel
+       Source0:        https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-4.20.5.tar.xz
+       Source1:        config
+       Source2:        cmdline
+
+   .. note::
+      TIP: Consider changing the Name in the RPM spec file to easily identify a modified kernel.
```
tpleavitt: I'm assuming you are referring the Name: field in the local copy of linux.spec shown above.  Any hard and fast rules to follow on the name?

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Including where to find valid kernel versions makes a lot of sense. 

I was hoping to using the `emphasize-lines` option in the [code-block directive](http://www.sphinx-doc.org/en/1.8/usage/restructuredtext/directives.html#directive-code-block ) to make it more apparent which lines to change but couldn't get it quite working. 

I'm hoping @mvincerx and team can help with a solution after the tech review is complete.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Right. Can you think of a way to make this more...clear?

No hard and fast rules that I am aware of. I imagine the limiting factor would be the filesystem max length filename.

source/clear-linux/guides/maintenance/autospec.rst
```diff
@@ -23,8 +23,10 @@ This guide requires that you:
 
 .. _install-tooling:
 
-Install the |CL| tooling framework
-==================================
+Install the |CL| development tooling framework
```
krisdale: Simplify section title to remove "Clear Linux OS" e.g. "Install development tooling framework"

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+If your change request is accepted, you do not need to maintain your own 
+modified kernel.
+
+Make enhancement requests to the |CL| distribution `on GitHub`_ .
+
+
+Customize the Linux kernel source
+=================================
+
+In some cases, it may be necessary modify the kernel to suit your specific
```
bktan8: add "to" before "modify"

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+
+.. include:: autospec.rst
+   :start-after: install-tooling-after-header:
+   :end-before: install-tooling-end:
+
+
+
+Clone the Linux kernel package 
+------------------------------
+Clone the existing kernel package re-pository from |CL| as a starting point.
```
bktan8: remove hyphen from "re-pository"

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+-------------------------
+
+|CL| tends to use the latest kernel available from `kernel.org`_, the Linux 
+upstream. The kernel version that will be built can be changed in the 
+RPM SPEC file.
+
+.. note:: 
+   While most packages in Clear Linux are typically packaged using 
+   :ref:`autospec`, the kernel is not. This means control files provided 
+   by autospec are not available  and changes must be made directly.
```
bktan8: Maybe "manually" would be better than "directly"?

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+
+* Disable Secure Boot
+* Maintain any updates to the kernel going forward
+
+To customize the kernel you will need a |CL| development environment, 
+make changes to the kernel, build the kernel, and install it.
+
+.. note::
+
+      This document shows how to obtain and compile a Linux kernel source
```
bktan8: This line should probably be part of the introduction to this document and doesn't need to be a Note.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+
+To customize the kernel you will need a |CL| development environment, 
+make changes to the kernel, build the kernel, and install it.
+
+.. note::
+
+      This document shows how to obtain and compile a Linux kernel source 
+      using the |CL| development tooling.
+
+      Source RPM files (SRPM) are available for all |CL| kernels, and can be used for development instead.
```
bktan8: I think this section on working with SRPM is not needed.  If we're going to mention it, then we should show how.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+      make clone_linux
+
+
+#. Navigate into the cloned package directory.
+
+   .. code-block:: bash
+
+      cd ~/clearlinux/packages/linux
+
+.. note::
```
bktan8: Maybe try to weave this Note into the introduction?  IMO, there are a lot of "Notes" in this document.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+      $EDITOR config-fragment
+
+   .. note::
+      Due to how |CL| packaging tools make use of :command:`mock` environments 
+      psuedo-GUI tools that abstract kernel configuration such as menuconfig 
+      do not work.
+
+
+#. Find the configuration values you are looking for. 
+   If a particular setting does not already exist , it can be added manually.
```
bktan8: remove space after "exist"

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,365 @@
...
+
+   For example, BTRFS support configuration is below indicating it is
```
bktan8: "For example, the snippet below shows how to enable BTRFS."

source/clear-linux/guides/maintenance/autospec.rst
puneetse: I added this to be more specific because the tooling is created by CL and specific to CL.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: I think it's worth mentioning. Maybe it merits another document in the future but shouldn't be a blocker.

Looking back at mailing list and IRC chats on this topic, the answer to this often mentions the SRPM availability. If a reader is familiar with SRPMs for kernels, they should know they're available in this context. 

I have moved this to the introduction per your other suggestion.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Agreed, there too many notes. There are a lots of spots where additional information/context is needed.  This is one of the ones I think is important to be close to the decision making point.

However, I have reduced the use of notes in other areas.

puneetse:
> @puneetse. I apologize that I missed this before. On line 351, could you please revise to make it a header, like so?
> 
> ```
> Related topics
> ***************
> ```
@mvincerx I tried that originally but it keeps complaining that the title level is inconsistent and not rendering properly on my local build. Can you give it a try?

mvincerx:
@puneetse.  I can change this in a follow up PR, or if you want to learn...

The error is that the second header of this document needs to be changed from: 
```
Request the change be included with the |CL| kernel
==================================
```
to: 
```
Request the change be included with the |CL| kernel
*******************************************************
```
Then *every header thereafter* needs to be changed to follow a cascading order per:
`Headings` :
`https://clearlinux.org/documentation/clear-linux/reference/collaboration/structure-formatting

puneetse:
>
> The error is that the second header of this document needs to be changed from:
> 

@mvincerx  thanks and good catch! I pushed the fix.

mvincerx:
@puneetse. Add an entry for `kernel-development`  after `kernel-modules`, in this doc : 
`clear-linux-documentation/source/clear-linux/guides/maintenance/maintenance.rst`

This is required for a proper build in Sphinx.

puneetse:
@mvincerx this should be in place already, see https://github.com/clearlinux/clear-linux-documentation/pull/383/files#diff-addad33a799a5169eb5a2274ab9d2b25R19

mvincerx:
@puneetse. You're right.  My error. (I thought I requested that before). You PR builds fine in testing now. 
LGTM.

mvincerx:
@bwarden, @miguelinux: Could *one* of you please review and validate this?

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
+.. _kernel-development:
+
+Kernel development on Clear Linux 
+#################################
+
+This document shows how to obtain and compile a Linux kernel source 
+using the |CL-ATTR| development tooling.
+
+The `kernels available in Clear Linux`_ aim to be performant and practical. 
+In some cases, it may be necessary modify the kernel to suit your specific
```
bwarden:
```suggestion
In some cases, it may be necessary to modify the kernel to suit your specific 
```

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+disabled in the kernel configuration file, :file:`.config` , at compile time.
+
+To manage unique changes that have been made to the kernel config file, 
+|CL| uses a kernel config fragment file named :file:`config-fragment`. 
+Managing kernel configuration changes with a configuration fragment file 
+instead of directly editing the :file:`.config` file helps identify the 
+unique configuration changes that have been made and makes applying any 
+future default configuration values easier. 
+
+The :file:`config-fragment` is the **only** file that is modified and is
```
bwarden: If they really want, they can modify the ```config``` file directly instead of messing with the fragments. It's really intended as an internal tool to allow us to manage mandatory kernel options across the multiple kernels we package.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+future default configuration values easier. 
+
+The :file:`config-fragment` is the **only** file that is modified and is 
+eventually merged with the main :file:`.config`.
+
+
+
+#. Open the kernel :file:`config-fragment` file in an editor.
+   
+   Due to how |CL| packaging tools make use of :command:`mock` environments
```
bwarden: They can use menuconfig if they work outside of mock, that is, directly in ```/var/lib/mock/clear-linux/root/builddir/build/BUILD/kernel-*```

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+
+   .. code-block:: bash
+      
+      # CONFIG_BTRFS_FS is not set
+      # CONFIG_BTRFS_FS_POSIX_ACL is not set
+
+#. Commit and save the changes to the :file:`config-fragment` file.
+
+
+#. Run the :command:`make config` command to apply the changes made in the
```
bwarden: May want to note that this action pulls in configuration fragments from the `kernel-config` package, and sandwiches the options in `config-fragment` between `base-X` and `mandatory-X`

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+
+
+#. Commit and save the changes to the RPM SPEC file.
+
+
+
+Build and install the kernel
+****************************
+After changes have been made to the kernel SPEC file and config file, 
+the kernel is ready to be compiled and packaged into a RPM.
```
bwarden:
```suggestion
the kernel is ready to be compiled and packaged into an RPM.
```

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+
+#. Start the compilation process by issuing the :command:`make build` command. 
+   This process is typically resource intensive and will take a while.
+
+   .. code-block:: bash
+      
+      make build
+
+
+#. The result there will be multiple :file:`.rpm` files in the :file:`rpms`
```
bwarden: Needs to be reworded.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,369 @@
...
+Alternatively, the kernel RPM bundle can be installed manually on a local 
+machine for testing. This approach works well for individual development or 
+testing. For a more scalable and customizable approach, consider using the 
+`mixer tool`_ to provide a custom kernel with updates.
+
+1. Install the kernel RPM onto the local system with the :command:`rpm` command.
+
+   .. code-block:: bash
+
+      sudo rpm -ivh linux-custom.<version>.<release>.x86_64.rpm
```
bwarden: Might be better to provide the rpm2cpio workflow instead; this may otherwise confuse people into thinking we're RPM-based.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: That makes sense technically. Is entering into the mock environment a flow that we want to support/encourage?

mvincerx:
@puneetse This PR shows a clean build in Sphinx.  👍  Just checking in. What's the status of this PR?

puneetse:
@mvincerx thanks for following up. I need some time to make some changes including feedback from @bwarden

mvincerx:
@puneetse Sounds good. Thanks for the update.

mvincerx:
Hi @puneetse. In one of your next commit messages, will you please add the following text? 
`Closes #249`
This will cause the issue to close upon my merge of this PR. 
https://github.com/clearlinux/clear-linux-documentation/issues/249

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Agreed and done.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: I've revamped this section to use a more standard kernel config editing workflow and not force the reader to use the CL-way workflow. 

I think the CL team's method of using kernel fragments is still useful for others to know. Perhaps we can publish a blog post on the subject and reference it here in the future.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: I've revamped this section to use a more standard kernel config editing workflow and not force the reader to use the CL-way workflow. 

I think the CL team's method of using kernel fragments is still useful for others to know. Perhaps we can publish a blog post on the subject and reference it here in the future.

puneetse:
@mvincerx @bwarden can you do a second pass on this? I've made changes since the last review and am ready to publish if you agree.

mvincerx:
@puneetse, Your PR has a clean build in Sphinx. 👍  I have several minor edits to make, which I'd like to do in a follow-up PR. First, can you make a few changes? 

- In title, replace "Clear Linux" with `|CL-ATTR|`
- On `:contents:` directive, change value from 2 to 1
- In "Clone the kernel package," can you please clearly identify the source? Slight disconnect here. 
Is it assumed that the user has this already?  If so, we should add a Prerequisites section below the `:contents:` in .rst file. You may find a point of entry here: https://clearlinux.org/documentation/clear-linux/guides/maintenance/kernel-modules

After these are done, I can merge this PR.

mvincerx:
@puneetse, Thanks for clarifying the URL to GH. LGTM. Ready to merge after @bwarden approves.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+
+
+Related topics
+**************
+
+* :ref:`kernel-modules`
+* :ref:`mixer`
+
+.. _`kernels available in Clear Linux`: https://clearlinux.org/documentation/clear-linux/reference/compatible-kernels
+.. _`on GitHub`: https://github.com/clearlinux/distribution
```
miguelinux: Here it is better to use

https://github.com/clearlinux/distribution/issues/new/choose

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+
+
+Customize the Linux kernel source
+*********************************
+
+After the kernel sources have been obtained, customizations to the kernel 
+configuration or source code can be made for inclusion with the kernel 
+build.
+
+These customizations are option optional.
```
bwarden:
```suggestion
These customizations are optional.
```

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+#. Commit and save the changes to the RPM SPEC file.
+
+
+
+Modify kernel boot parameters
+=============================
+The kernel boot options are passed from the bootloader to the kernel with 
+command-line parameters. 
+
+While temporarily changes can be made to kernel parameters on a running
```
bwarden:
```suggestion
While temporary changes can be made to kernel parameters on a running 
```

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+
+
+
+Modify kernel boot parameters
+=============================
+The kernel boot options are passed from the bootloader to the kernel with 
+command-line parameters. 
+
+While temporarily changes can be made to kernel parameters on a running 
+system or on a during boot, you can also modify the default parameters that
```
bwarden: May want to point to `man clr-boot-manager` for info on customizing the command line on a running system.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+
+
+Build and install the kernel
+****************************
+After changes have been made to the kernel source and RPM SPEC file, 
+the kernel is ready to be compiled and packaged into an RPM.
+
+#. Start the compilation process by issuing the :command:`make build` command. 
+   This process is typically resource intensive and will take a while.
+
```
bwarden: Might be good to add a note about enabling ccache plugin for mock to speed up rebuilds.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+machine for testing. This approach works well for individual development or 
+testing. For a more scalable and customizable approach, consider using the 
+`mixer tool`_ to provide a custom kernel with updates.
+
+1. Install the kernel onto the local system by extracting the RPM with the 
+   :command:`rpm2cpio` command.
+
+   .. code-block:: bash
+
+      sudo rpm2cpio linux<NAME>-<VERSION>-<RELEASE>.x86_64.rpm | (cd /; sudo cpio -i -d -u -v);
```
bwarden: Don't need the `sudo` on `rpm2cpio` command; just the one for `cpio`.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Since the primary message is about changing the compiled cmdline, I think it's better not to elaborate on the other ways.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+You will need to make modifications to the :file:`.config` file, and include 
+it in the kernel package. 
+
+
+#. Make sure you have followed the steps to 
+   `Pull a copy of the Linux kernel source code`_ and are in the kernel 
+   source working directory.
+
+
+#. If you have an existing :file:`.config` file from an old kernel, copy it
```
miguelinux: Why does the user need to generate a new config and not use the current one from Clear Linux?

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+   `Pull a copy of the Linux kernel source code`_ and are in the kernel 
+   source working directory.
+
+
+#. Make any desired code changes to the Linux source code files.
+
+
+#. Track and commit your changes to the local git repo. 
+
+   .. code-block:: bash
```
miguelinux: As the current instruction the following command will fail as it is not a git repository.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: I had it this way originally but changed it to not force readers to use the kernel fragmenting w/ CL tooling based on a comment from @bwarden. I think kernel fragment worflow (pulling base+mandatory from kernel-fagment) would be a good topic for a blog post and link this article from.

In the mean time do you think it's a good idea to do a `cp ../config .config` into the kernel source code directory so that they're using the config with packaging as the base config instead of `make defconfig`?

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,462 @@
...
+      $EDITOR linux.spec
+
+
+#. Locate the section of the SPEC file that contains existing patch 
+   variable definitions and append your patch file name. Ensure the 
+   patch number does not collide with an existing patch. 
+   In this example, the patch file is called 
+   :file:`2001-my-patch-for-driver-A.patch`
+
+   .. code-block:: bash
```
miguelinux: There is no mention on the document to apply the current CL patch set

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: It did not fail for me when I did it. It seems to do a `git init` on the local directory internally. It adds all files as new, and then we tell them only to stage/add the only the changed files

Did you try it and run into issue or can you recommend a better workflow that is both easy to read and execute?

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Can you clarify what you mean? In this example, readers would be using the existing RPM SPEC file which already has the CL patches. This is just a snippet.

source/clear-linux/guides/maintenance/kernel-development.rst
miguelinux: agree

source/clear-linux/guides/maintenance/kernel-development.rst
bwarden: > do you think it's a good idea to do a `cp ../config .config` into the

Yes.

source/clear-linux/guides/maintenance/kernel-development.rst
miguelinux: Yes, but the user did not apply any of those patches to the kernel source.

They should:

1. get the repo
2. get the kernel sources
3. apply CL patches to the kernel sources
4. copy `config` to kernel sources as `.config`
5. do development (any changes)
6. create the patch
7. add to the repo/spec file

miguelinux: 8. copy the modifed `.config` to repo `config`

miguelinux:
> I had it this way originally but changed it to not force readers to use the kernel fragmenting w/ CL tooling based on a comment from @bwarden. I think kernel fragment worflow (pulling base+mandatory from kernel-fagment) would be a good topic for a blog post and link this article from.
> 
> In the mean time do you think it's a good idea to do a `cp ../config .config` into the kernel source code directory so that they're using the config with packaging as the base config instead of `make defconfig`?
> 

Yes, take config as base.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: Thanks @miguelinux. As far as I can tell, only step 3 is missing to apply the CL patches to the copy of the kernel source directory. Is this required? If so, can you help explain why? Is it so that any user-generated patch files take into the account the CL patches?

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: @miguelinux in any case, I have added an explicit `git init` and initial `git add -A` command to be explicit. Let me know if you think this is a good approach.

source/clear-linux/guides/maintenance/kernel-development.rst
puneetse: @miguelinux in any case, I have added step 3. Let me know if you think it looks good.

miguelinux: if you want to modify the kernel from CL you need to have All the patches applied to the current kernel.
Some patches add Kconfig options, so when you use the `config` file as reference you could lose those features. 

If the kernel won't be based on CL so you don't need to apply the patches or use the config as base.

source/clear-linux/guides/maintenance/kernel-development.rst
```diff
@@ -0,0 +1,490 @@
...
+
+#. Make sure you have followed the steps to 
+   `Pull a copy of the Linux kernel source code`_ and are in the kernel 
+   source working directory.
+
+
+#. Initialize the kernel source directory as a new git repo and create a 
+   commit with all the existing source files to begin tracking changes. 
+
+   .. code-block:: bash
```
miguelinux: like it :+1:

source/clear-linux/guides/maintenance/kernel-development.rst
miguelinux: :+1:

mvincerx:
@puneetse.  Clean build in Sphinx. 👍 Compliments on your perseverance. Like we discussed, I'll create a followup PR for documentation cleanup with you as Reviewer. LGTM. Merging.

PR #396, opened 2019-02-11
Fix `dd` arg error

shrmrf:
`4M` instead of `4m`

### DIFF
--- a/source/clear-linux/get-started/bootable-usb/bootable-usb-beta-all.rst
+++ b/source/clear-linux/get-started/bootable-usb/bootable-usb-beta-all.rst
@@ -137,7 +137,7 @@ Burn the |CL| image onto a USB drive
 
    .. code-block:: console
 
-      sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4m
+      sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4M
 
 
    Adding an ‘r’ in front of the disk identifier should help speed up the


mvincerx:
Thanks, as always for contributing @shrmrf.   I hesitate to call this an error. From what I've found, the "success" of lower or uppercase usage can be hardware dependent, but `1M` is probably faster.   It may be the case that MacOS changed/ allowed different numerals (e.g., M, m, k, etc) from one OS version to the next. 

This doc may be improved if we add a note.  But first, I have a few questions.

* Did you experience a faster rate by using the uppercase `bs=4M` (vs. `bs=4m`) ?
* Can you share which mac and OS that you tested this on? 

I tested the dd command using `bs=4m`, and it burned an image fine --and I booted from it. 

TEST
I used an (old!) MacBook Pro 2012 with OS X El Capitan, 10.11.6. 
*  `sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4m` . This worked fine (but was slow)
* `sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4M`  This failed with error message: `illegal numeric value`

Look forward to hearing more about your test.

shrmrf:
Hey @mvincerx 

>I used an (old!) MacBook Pro 2012 with OS X El Capitan, 10.11.6.

>    * `sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4m` . This worked fine (but was slow)

>    * `sudo dd if=./clear-[version number]-[image type] of=/dev/rdisk2 bs=4M`  This failed with error message: `illegal numeric value`


>Look forward to hearing more about your test.

Actually, I did _not_ test this on a Mac at all. I just assumed that since Linux uses the capital `M`, it's the same for Mac. But I stand corrected.

But then I did some StackOverflow search after reading your message and found this: https://superuser.com/a/1346063 .. apparently `rdisk2` is slower sometimes..?

Sorry for submitting the issue without verification. Closing.

mvincerx:
Hey @shrmrf.  Thanks, anyway, for checking the repo! We always appreciate contributions.

PR #421, opened 2019-03-09
kvm: Configure and start SSH server

anselmolsm:
Step 1 actually is a configuration step, it does not enable SSH. Besides that, the server has to be started.


mvincerx:
LGTM. Thanks @anselmolsm.

PR #422, opened 2019-03-09
kvm: Simplify step to install vncviewer

anselmolsm:
swupd bundle-add tigervnc is enough to install a vncviewer (5MB vs 125 MB of desktop-apps-extras)


mvincerx:
LGTM.

PR #431, opened 2019-03-16
Adds missing `_` leading space before file name; root cause image failure

mvincerx:
Rushing this PR out. FYI only. 1 char fix. 


mvincerx:
For the record, this was not root cause of the error. Root cause determined by ODX as lack of server space.

PR #436, opened 2019-03-28
Bootable USB

jairglez:
Update the 'Create a bootable USB drive' manual with an additional caution to prevent users from errors that may occur when the USB drive partitions are not properly unmounted before burning an image, and allow them to correct the issues if they occur.

Also, update the example burning command to improve its performance and progress output.


mvincerx:
LGTM.
Grazie per l'aiuto Beth. In bocca al lupo!
Thanks, @jairglez.

PR #437, opened 2019-03-29
Update to parse-link-check.py

intelkevinputnam:
Updated parse-link-check.py script to ignore anchors. 

Motivation:

1. Anchors in .rst files in GitHub are considered broken by the tester though they resolve correctly when manually clicking through on the browser.
2. Broken anchors should not block the site from being built.
3. There are now 3 categories in broken_links.html
   1. unmatched broken links
   2. broken links on the white list
   3. broken anchors


mvincerx:
What do you use as your input file, as argument, for testing?

mvincerx:
Let's go over this PR on Monday when we're both in the office. When I run `make link`, I notice several build errors, but it's difficult to distinguish between those errors that you wish to raise vs. those that could be legitimate build errors.   

Maybe this can be helped by revising the embedded documentation in the .py file. Let's discuss.

intelkevinputnam:
@mvincerx Please review again.

mvincerx:
LGTM. Thanks for improving documentation of your code.

PR #440, opened 2019-04-03
Remove optional curl method to download installer img.

mvincerx:
- Discontinue support for older txt file and for embedded parsing commands.


mvincerx:
@phmccarty, @mbelluzzo,  please review if time. I discussed this with @mbelluzzo, and we agreed that while adding a new ` | grep` option may work, it's not worth the technical debt. Since this step was optional,  I see no reason to include it.

mbelluzzo:
:+1:

mvincerx:
Thanks, @mbelluzzo.

Issue #419, opened 2019-03-09
kvm: OVMF file and steps 3 and 4


anselmolsm:
**Describe the error/improvement to an existing document or image**

In the [guide to run Clear Linux as a KVM guest OS](https://clearlinux.org/documentation/clear-linux/get-started/virtual-machine-install/kvm), section "Download and launch the virtual machine":
* Step 3 tells users to download the OVMF file (side note: there could be a copy&paste friendly curl line to download it);
* Step 4 tells user to do a cp of the same file from /usr/share/qemu.

These steps seem to conflict (imagine if the user downloads the file to workdir, then follow step 4... one file would overwrite the other).

Considering the recommendation in the "Note" after those steps, it seems steps 3 and 4 target different users: step 3 non-CL hosts, step 4 CL hosts.

Still, I downloaded OVMF from the link provided then compared (md5sum) that file with the one at /usr/share/qemu in my Clear Linux host, and they are different... So that made me think: Which version is the best to use?

**Environment (please complete the following):**
 - Clear Linux OS version: 28230

PR #447, opened 2019-04-06
kvm: Revise conflict between steps 3 and 4.

mvincerx:

### DIFF
--- a/source/clear-linux/get-started/virtual-machine-install/kvm.rst
+++ b/source/clear-linux/get-started/virtual-machine-install/kvm.rst
@@ -28,7 +28,7 @@ Install QEMU-KVM
 
         swupd bundle-add kvm-host
 
-    * |CL| option: Add this bundle to lanch the GUI upon boot.
+   * |CL| option: Add this bundle to lanch the GUI upon boot.
 
      .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


source/clear-linux/get-started/virtual-machine-install/kvm.rst
```diff
@@ -76,13 +76,13 @@ Download and launch the virtual machine
```
mvincerx: @eadamsintel. I need your opinion. A user raises a good point: 
The current steps 3-4 conflict, under **Download and launch the virtual machine**.
https://clearlinux.org/documentation/clear-linux/get-started/virtual-machine-install/kvm#kvm

Step 3 asks a user to download the OVMF.fd from image directory (but doesn't explain where or how to use it).
Step 4 asks a user to copy an existing OVMF.fd from directory here: `cp /usr/share/qemu/OVMF.fd`

These are clearly two different methods, and hence, confusing. In my revision, instead, I have the user copy the OVMF.fd from the Downloads dir, keeping with the intent of Step 3. 

Please make a recommendation, or best practice, as to which step, 3 or 4, should be followed. 
I will delete the step that isn't the best practice.  
See also:  https://github.com/clearlinux/clear-linux-documentation/issues/419

eadamsintel: I'm not sure.  The user said the ovmf files are different so that issue needs to be resolved first. Are they diff because they come from different clear linux releases and it got updated or are they different because a timestamp is part of the binary thus changing every file. 

I would just put a curl command to point to https://cdn.download.clearlinux.org/image/OVMF.fd since that location is not changing. It would cover both cases of using clear linux or using another host.

mvincerx:
@eadamsintel. Please see my last commit and let me know if that works.

eadamsintel:
> @eadamsintel. Please see my last commit and let me know if that works.

LGTM

PR #459, opened 2019-04-15
kata: removes configuration step /etc/systemd/system/docker.service.d/50-runtime .conf

### LINKED ISSUE #453: Kata containers setup guide has extra unnecessary steps
https://clearlinux.org/documentation/clear-linux/tutorials/kata

Sometime in 2018 Clear Linux was updated so that the /etc/systemd/system/docker.service.d/50-runtime.conf is automatically created when you "sudo systemctl start docker" for the first time and that automatically creates the default runtime for Kata.  This information is still useful to have in the tutorial as a reference but it is automatically done and might be confusing.

 - Clear Linux OS version: 28300 but also many older and newer versions

mvincerx:
- 50-runtime.conf file automatically created on CL builds > 27000
- Adds Troubleshooting, and adds prev manual conf option
- Closes #453

### DIFF
--- a/source/clear-linux/tutorials/kata.rst
+++ b/source/clear-linux/tutorials/kata.rst
@@ -110,7 +110,7 @@ To do so, enter:
 .. code-block:: bash
 
    sudo mkdir -p /etc/systemd/system/docker.service.d/
-   cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/51-runtime.conf
+   cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/50-runtime.conf
    [Service]
    Environment="DOCKER_DEFAULT_RUNTIME=--default-runtime kata-runtime"
    EOF
--- a/source/clear-linux/tutorials/kata.rst
+++ b/source/clear-linux/tutorials/kata.rst
@@ -99,7 +99,7 @@ To check which runtime your system uses, enter:
 
 .. note::
 
-   To change the Docker devicemapper settings for production use, see
+   To change the Docker storage driver, see
    :ref:`additional-docker-configuration`.
 
 For some |CL| versions before 27000, you may need to manually
--- a/source/clear-linux/tutorials/kata.rst
+++ b/source/clear-linux/tutorials/kata.rst
@@ -90,36 +90,27 @@ it automatically uses the runtime supported by the system.
 Troubleshooting
 ===============
 
-To check which runtime your system uses, enter:
+- To change the Docker storage driver, see
+  :ref:`additional-docker-configuration`.
 
-.. code-block:: bash
+- For some |CL| versions before 27000, you may need to manually
+  configure Docker\* to use Kata Containers by default.
 
-   sudo docker info | grep runtime
+  To do so, enter:
 
+  .. code-block:: bash
 
-.. note::
+     sudo mkdir -p /etc/systemd/system/docker.service.d/
+     cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/50-runtime.conf
+     [Service]
+     Environment="DOCKER_DEFAULT_RUNTIME=--default-runtime kata-runtime"
+     EOF
 
-   To change the Docker storage driver, see
-   :ref:`additional-docker-configuration`.
+- To check the version of |CL| on your system, enter:
 
-For some |CL| versions before 27000, you may need to manually
-configure Docker\* to use Kata Containers by default.
+  .. code-block:: bash
 
-To do so, enter:
-
-.. code-block:: bash
-
-   sudo mkdir -p /etc/systemd/system/docker.service.d/
-   cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/50-runtime.conf
-   [Service]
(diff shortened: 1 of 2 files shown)


source/clear-linux/tutorials/kata.rst
```diff
@@ -97,10 +87,32 @@ environment:
...
+To do so, enter:
+
+.. code-block:: bash
+
+   sudo mkdir -p /etc/systemd/system/docker.service.d/
+   cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/51-runtime.conf
+   [Service]
+   Environment="DOCKER_DEFAULT_RUNTIME=--default-runtime kata-runtime"
+   EOF
+
```
mvincerx: @eadamsintel. Why does the older version use `51-runtime.conf`, but the new version uses `50-runtime.conf`? 
Should this older version be corrected to `...50-runtime.conf`?

source/clear-linux/tutorials/kata.rst
```diff
@@ -97,10 +87,32 @@ environment:
 You do not need to manually configure the runtime for Docker, because
 it automatically uses the runtime supported by the system.
 
-Check which runtime your system is using with the command:
+Troubleshooting
+===============
+
+To check which runtime your system uses, enter:
```
mvincerx: In validating this command, I observe (on 28330) an error occurs: 

`WARNING: devicemapper: usage of loopback devices is strongly discourage for production use. 
      use `--storage-opt dm.thinpooldev` to specify a custom block storage device`

How would you suggest addressing this scenario? 

Note: When I run a typical `sudo docker info`, output is as expected.

source/clear-linux/tutorials/kata.rst
eadamsintel: Sure.  You can manually create any version you want.  I'd make it consistent.

source/clear-linux/tutorials/kata.rst
eadamsintel: The default should be to use overlay.  https://docs.docker.com/storage/storagedriver/device-mapper-driver/  Docker is deprecating devicemapper support anyways.  You must have ```
{
  "storage-driver": "devicemapper"
} 
``` set in a config somewhere.

source/clear-linux/tutorials/kata.rst
mvincerx: Agreed. I'll make them both `...50-runtime.conf`

source/clear-linux/tutorials/kata.rst
mvincerx: Yes, this is what I had: 
{
 `storage-driver": "devicemapper"
}

Docker recommends using `overlay2` instead of `overlay`.  Can you please advise? 
If we change to either of these, I'll revise our Docker tut to reflect that change too (where I took it from):
https://clearlinux.org/documentation/clear-linux/tutorials/docker#id4

eadamsintel: I would follow Docker's recommendation.

mvincerx: Please review my latest commit, above, for added note in docker.rst, and adding a link to it from kata.rst to docker.rst.

eadamsintel: Have you tested the daemon.json change and adding devicemapper?  I was trying out firecracker today with Kata which requires devicemapper and just adding this one setting prevented docker from restarting.  The issue was that if  /usr/lib/systemd/system/docker.service.d/clearlinux.conf had overlay2 specified and I added this daemon.json to /etc/docker with devicemapper then Docker failed to start.   If I changed /usr/lib/systemd/system/docker.service.d/clearlinux.conf to be devicemapper that worked and also worked if I had the same setting in that location and /etc/docker/daemon.json.   I didn't have time to look into this further to test and see if this is a Clear Linux bug or a misunderstanding in how to properly override files.  I think the proper way to override is in /etc/system/systemd/docker.service.d/<some conf> but that didn't work for me either or I had my file formatted wrong.

mvincerx: Please see my latest commit, and let me know if you find this acceptable. 
I changed the storage driver in `etc/docker/daemon.json`  and it worked fine for docker.rst and kata.rst in this PR. I can't speak to firecracker. 

In kata.rst, this command produced no stdout when I ran it: `docker info | grep runtime`
I suggest that we remove it, unless you have an alternative suggestion.

eadamsintel: That works for me.  There might be an underlying issue when you try to change to devicemapper through this method but what you suggest should work. Devicemapper is being deprecated in Docker anyways.

mvincerx: In response to your concern about devicemapper, in kata.rst, I added a link back to docker.rst, where I recommend using `overlay2`. There, I also provide links to Docker docs if users want to use other storage drivers. 

Will you please give a Y/N vote on this PR?

eadamsintel: Approved and LGTM

PR #460, opened 2019-04-15
yubikey: Add sudo in second command to reload udev rules.

### LINKED ISSUE #458: Wrong shell command in YubiKey U2F support tutorial
**Error**

The 4th bullet point in section `Enable Linux udev rules for YubiKey` has a shell command, such as
```bash
sudo udevadm control --reload-rules && udevadm trigger
```
This won't work as the second statement `udevadm trigger` also needs root privilege. The correct one shall be
```bash
sudo udevadm control --reload-rules && sudo udevadm trigger
```

mvincerx:
- Closes #458


mvincerx:
Cheers.

PR #466, opened 2019-04-25
Update command to verify CL root partition in fix-broken-install.rst 

puneetse:
This PR:

- Fixes the command in fix-broken-install.rst to verify the mounted root partition is a Clear Linux partition. The current command is incorrect because it checks a file from the temporarily booted live OS, not the mounted partition to repair. 

- Minor elaboration on mounting step for clarity.


mvincerx:
LGTM.

PR #467, opened 2019-04-26
Revise config media installation; add adv configuration option

mvincerx:


mvincerx:
@mdhorn or reagan (gh handle?) will you please review as soon as possible? This is hot. Thanks.

mdhorn:
One point, there has not been a new release of the Live Desktop Beta image since 28900
see https://download.clearlinux.org/image/
which means the desktop beta image visible to the public still has the TUI 1.1.8

The sever (installer image or installer ISO) have been updated to the latest release and this new version of the TUI 1.2.1 (1.2.2 as of Friday).

mdhorn:
Figure 7 has a screen show safe media selected, but nothing in the chooser box and a warning that it will be erased -- this combination shouldn't be possible and is probably a bug. Regardless, we should not show the screen in that undesired state.
image

mdhorn:
After the "Destructive Installation" section, we should address the use of the "Enable Encryption" checkbox. Checking the box will prompt for a LUKS encryption password (and confirmation).
This currently will only encrypt the root '/' partition.

mvincerx:
@mdhorn I removed the Warning message from Fig 7 that apparently reflects a bug.

mvincerx:
@mdhorn So I revised lines 377-379 on encryption, making it only available for `/boot` partition. That's a change that I wasn't aware of. To keep consistent with the latter, I removed the note that previously started on 397. Please confirm that this is what you want. 



> which means the desktop beta image visible to the public still has the TUI 1.1.8
We should discuss a strategy for this on Monday 4/29.  It may complicate the docs--if only temporarily.

mdhorn:
> @mdhorn So I revised lines 377-379 on encryption, making it only available for `/boot` partition. That's a change that I wasn't aware of. To keep consistent with the latter, I removed the note that previously started on 397. Please confirm that this is what you want.
> 
> > which means the desktop beta image visible to the public still has the TUI 1.1.8
> > We should discuss a strategy for this on Monday 4/29.  It may complicate the docs--if only temporarily.


Encryption can be enabled for any partition expect `/boot` under the Advanced Configuration.
But, if you Check the Encryption checkbox on the Media Configuration main page, that will only enable encryption on the `/` root partition using the defaults of "Safe" or "Destructive".

mvincerx:
Hi @mdhorn. Please see my latest commit per our discussion this morning. Hoping to merge this today.

mdhorn:
@mvincerx LGTM 👍

PR #471, opened 2019-05-03
Fixed incorrectly formed link that was resulting in 404

intelkevinputnam:
Found it during link checking. The link is formatted as a URL reference but needed to use :ref: for internal link.


mvincerx:
LGTM

PR #474, opened 2019-05-07
Update NVIDIA driver documentation to include openGL support

puneetse:
This PR:

-  Updates the NVIDIA installer command to isolate its files under `/opt/nvidia/` and

- Enables inclusion of OpenGL support. https://github.com/clearlinux/distribution/issues/324

- Enables inclusion of compat32 libs (needed for Steam)

- Simplifies the process by using the .run with switches directly file instead
of manually extracting the files and `--silent` switch, avoiding an intermediary steps. 

- Makes using the LTS kernel more of a suggestion than a requirement. 

- Adds a section for tips on debugging.


source/clear-linux/tutorials/nvidia.rst
```diff
@@ -139,11 +160,18 @@ Install the NVIDIA Linux Driver
...
+      sudo swupd verify --quick --fix --bundles=lib-opengl
+
 .. note::
 
-   The NVIDIA installer places files under the :file:`/usr` subdirectory which
-   are not managed by |CL| updates. The :command:`swupd verify --fix` command
-   should be avoided with the proprietary NVIDIA drivers in use.
+   The NVIDIA installer places some files under the :file:`/usr` subdirectory
+   which are not managed by |CL| updates. The :command:`swupd verify --fix`
+   command should be avoided with the proprietary NVIDIA drivers in use.
```
bwarden: Might want to explain more here why we're running verify with --bundles above.

source/clear-linux/tutorials/nvidia.rst
```diff
@@ -152,19 +180,33 @@ Uninstalling the NVIDIA driver
 The NVIDIA drivers and associated software can be uninstalled and nouveau
 driver restored by: 
 
-#. Remove the previously created file :file:`etc/modprobe.d` that is
-   preventing nouveau from loading.
+#. Remove the previously created file :file:`/etc/modprobe.d` that is
```
bwarden:
```suggestion
#. Remove the previously created file :file:`/etc/modprobe.d` that
```

source/clear-linux/tutorials/nvidia.rst
puneetse: I think this line does that:
> Run a verification to restore system files that the NVIDIA installer likely
   deleted.

But I get your point - it's kind of conflicting that we say `swupd verify`, then you shouldn't `swupd verify`. I'll play with the wording a bit to see if it can be explained easier.

puneetse: I've pushed an update to the wording. Let me know what you think.

mvincerx:
LGTM. Thanks for hanging with the debate/dialogue all.

PR #476, opened 2019-05-08
live server: Creates bare-metal-install-server; replaces bare-metal-install.

mvincerx:
- Revise all submenus per release: clear-29180-0502-live-server.iso
- Revises User Manager to Manage User
- Revises majority of figures.


source/clear-linux/get-started/bare-metal-install-server/bare-metal-install-server.rst
```diff
@@ -0,0 +1,857 @@
...
+#. Follow your OS instructions to create a bootable USB drive.
+
+   * :ref:`bootable-usb`
+
+#. After downloading the image, verify and decompress the file per your OS.
+
+Install |CL| on your target system
+**********************************
+
+We formatted the previously created USB drive as a UEFI boot device. Our
```
mdhorn: Technically, the USB contains an ISO image that will boot with either UEFI or legacy BIOS, but you must enable UEFI to install from the live image.
I would drop the line "We formatted the previously created USB drive as a UEFI boot device."
You could say something like "Ensure your system is configured to boot UEFI."

mvincerx: Done. I revised that entire paragraph. It was a holdover from before my time. See latest commit.

PR #487, opened 2019-05-14
Revises Create a bootable USB drive on Windows, Rufus upgrade to 3.5

mvincerx:
- Revises 2 screenshots and adds note.


mvincerx:
@evil-brain will be main reviewer bc he helped on updating the iso. Please accept invitation to repo asap. 
@intelkevinputnam review is optional.

source/clear-linux/get-started/bootable-usb/bootable-usb.rst
```diff
@@ -185,12 +186,22 @@ Burn the |CL| image onto a USB drive
...
+   .. figure:: figures/bootable-usb-windows-02.png
       :scale: 80 %
       :alt: Rufus utility
 
       Figure 2: Rufus utility
 
+#. When the dialogue appears, select
+   :guilabel:`Write in ISO image mode (Recommended)`. See Figure 3.
+
+   .. figure:: figures/bootable-usb-windows-03.png
```
puneetse: Non blocking for this PR, but would like to share my thinking.

On Rufus, the .IMG defaulted to GPT partition scheme, UEFI only. 

With the switch to ISO, Rufus defaults to MBR partition scheme allowing for BIOS and UEFI booting. I'm thinking it would cause less errors/confusion if we explicitly say to select GPT in the documentation. That way, if someone tries to boot Clear on a legacy BIOS system it will fail faster (and often with a more obvious error message depending on the EFI).

mvincerx: @puneetse, Let me check the default menu options in Rufus when I return to the office.

source/clear-linux/get-started/bootable-usb/bootable-usb.rst
```diff
@@ -176,6 +176,7 @@ Burn the |CL| image onto a USB drive
    |CAUTION-BACKUP-USB|
 
 #. Download the `Rufus`_ utility to burn the image onto a USB drive.
+   **Use only the latest version of Rufus**.
```
mdhorn: I would suggest adding a comment of what the latest version is at the time of this publication; otherwise, you're committed to checking for the latest and any possible changes all of the time.

source/clear-linux/get-started/bootable-usb/bootable-usb.rst
evil-brain: The ISO actually supports Legacy BIOS firmware.
Using the default settings (IIRC, MBR), I have tested rufus on both legacy and uefi with success.

source/clear-linux/get-started/bootable-usb/bootable-usb.rst
evil-brain: I agree here, we should include which version we validated with.

mvincerx: I will add the version but retain the note. Users are on notice that they should use the latest Rufus.

mvincerx: See latest commit, @evil-brain.

mvincerx:
Thanks @evil-brain. Merging.

PR #494, opened 2019-05-17
Add introductory text and link to benchmark results

bd-dean:


mvincerx:
Nice addition and link. LGTM.

Issue #500, opened 2019-05-19
Update nvidia driver documentation with upgrade instructions


puneetse:
**Describe the error/improvement to an existing document or image**
https://clearlinux.org/documentation/clear-linux/tutorials/nvidia

- Update to include instructions on how to upgrade from one version of NVIDIA drivers to another (requires booting without desktop).

- Update to include `flatpak update` step after installation to pull the flatpak runtime corresponding to the NVIDIA driver for flatpak apps, otherwise they will not run after a driver update.

PR #501, opened 2019-05-20
Update NVIDIA doc to have update instructions

puneetse:
This PR:

- Updates the NVIDIA driver document to have a section on upgrading from one version to another
- Separates configuring dynamic link instructions from installation instructions
- Minimizes the use of "Linux" when referring to the NVIDIA drivers after the download section.

Closes https://github.com/clearlinux/clear-linux-documentation/issues/500


mvincerx:
Clean build in the doc set. LGTM.

PR #511, opened 2019-05-24
Fix license in autospec example.

puneetse:
The correct identifier for the Intel Open Source license is just "Intel" not "Intel Corporation"

### DIFF
--- a/source/clear-linux/guides/tooling/autospec.rst
+++ b/source/clear-linux/guides/tooling/autospec.rst
@@ -242,7 +242,7 @@ and dependencies in order for autospec to complete a successful build.
 
    .. code-block:: bash
 
-      echo "Intel" > opae-sdk.license
+      echo "MIT" > opae-sdk.license
 
 #. Run autospec again:
 
--- a/source/clear-linux/guides/tooling/autospec.rst
+++ b/source/clear-linux/guides/tooling/autospec.rst
@@ -242,7 +242,7 @@ and dependencies in order for autospec to complete a successful build.
 
    .. code-block:: bash
 
-      echo "MIT" > opae-sdk.license
+      echo "BSD-3-Clause MIT" > opae-sdk.license
 
 #. Run autospec again:
 
--- a/source/clear-linux/guides/tooling/autospec.rst
+++ b/source/clear-linux/guides/tooling/autospec.rst
@@ -242,7 +242,7 @@ and dependencies in order for autospec to complete a successful build.
 
    .. code-block:: bash
 
-      echo "Intel Corporation" > opae-sdk.license
+      echo "Intel" > opae-sdk.license
 
 #. Run autospec again:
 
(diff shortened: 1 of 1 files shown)


source/clear-linux/guides/tooling/autospec.rst
```diff
@@ -242,7 +242,7 @@ and dependencies in order for autospec to complete a successful build.
 
    .. code-block:: bash
 
-      echo "Intel Corporation" > opae-sdk.license
+      echo "Intel" > opae-sdk.license
```
bktan8: Please change it to "MIT" instead of "Intel" because it's using the MIT license.

ahkok:
`MIT` would also be a better example (since, much more common).

But we should really insert a reference to spdx.org/licenses here....!

source/clear-linux/guides/tooling/autospec.rst
phmccarty: @bktan8 I have found different information... From https://opae.github.io:

> The OPAE SDK source, unless otherwise noted, is released under a BSD 3-clause license

The [COPYING](https://github.com/OPAE/opae-sdk/blob/master/COPYING) file provides the full details. [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html) for almost all code. The one exception is the `safe_string` library included in the tree, which is [MIT](https://spdx.org/licenses/MIT.html) licensed.

So, my recommendation is:

```
echo "BSD-3-Clause MIT" > opae-sdk.license
```

source/clear-linux/guides/tooling/autospec.rst
```diff
@@ -238,11 +238,11 @@ and dependencies in order for autospec to complete a successful build.
 
       cd packages/opae-sdk
 
-#. Add a license:
+#. Add a valid license identifier from the `SPDX License List <https://spdx.org/licenses/>`_:
 
    .. code-block:: bash
 
-      echo "Intel Corporation" > opae-sdk.license
+      echo "MIT" > opae-sdk.license
```
phmccarty: Should be `BSD-3-Clause MIT`. See the [COPYING](https://github.com/OPAE/opae-sdk/blob/master/COPYING) file.

mvincerx: I'm engaging legal to review this as well.

bktan8: Thanks @phmccarty for verifying it's `BSD-3-Clause MIT`.

mvincerx:
@phmccarty, @ahkok, please let us know if you approve of the latest commit. 
@puneetse thanks for going the extra mile on details for multiple licenses.

mvincerx:
I wait until morning to see if @ahkok responds.

mvincerx:
@ahkok . Any comments? As long as I have approval from you or @phmccarty, I'd like to merge.

Issue #532, opened 2019-06-07
Update Clear Linux OS image type documentation


tpleavitt:
The document https://clearlinux.org/documentation/clear-linux/guides/maintenance/image-types#image-types is out of date and needs to be updated to include the new installer iso images.  I've made a similar request in https://github.com/clearlinux/distribution/issues/865

tpleavitt:
@intelkevinputnam , can you work with @gtkramer and make this a high priority.  He wants to get these updated this week if possible.

gtkramer:
source/clear-linux/guides/maintenance/validate-signatures.rst also needs updated to no longer reference the installer image.

gtkramer:
The following is the tentative list of correct image types that we officially publish and should have listed as image types in our documentation:
* aws
* azure
* azure-docker
* azure-machine-learning
* cloudguest
* gce
* hyperv
* kvm
* legacy-kvm
* live-desktop
* live-server
* pxe
* vmware

gtkramer:
The images link in https://clearlinux.org/documentation/clear-linux/guides/maintenance/image-types#image-types should point to https://clearlinux.org/downloads to avoid redirecting people to our file server.

gtkramer:
https://clearlinux.org/downloads will be updating in the next day or two here to expose the full list of images above, adding aws, pxe, and gce.

gtkramer:
README-IMAGES.html has been removed from the file server in favor of people going to https://clearlinux.org/downloads or visiting the public documentation for image types.

gtkramer:
@bktan8 fyi

intelkevinputnam:
@gtkramer Got it. Working up a PR now.

intelkevinputnam:
PR#548 opened to address this.  https://github.com/clearlinux/clear-linux-documentation/pull/548

krisdale:
Issue resolved with PR #548

PR #548, opened 2019-06-14
Update image types to match current offerings.

intelkevinputnam:
* updated image types to match current offerings
* fixed broken link in ister.rst - now pointing at image-types.rst

### DIFF
--- a/source/clear-linux/reference/image-types.rst
+++ b/source/clear-linux/reference/image-types.rst
@@ -60,9 +60,6 @@ Table 2 lists the currently available images that are platform specific.
    * - azure-machine-learning.vhd
      - Virtual Hard Disk for use on Microsoft Azure cloud platform with the `machine-learning-basic` bundle installed.
 
-   * - cloud.img
-     - Image for use by cloud deployments such as OpenStack\*.
-
    * - cloudguest.img
      - Image with generic cloud guest virtual machine (VM) requirements installed.
 
--- a/source/clear-linux/reference/image-types.rst
+++ b/source/clear-linux/reference/image-types.rst
@@ -35,14 +35,11 @@ Table 2 lists the currently available images that are platform specific.
    * - Image Type
      - Description
 
-   * - installer.img
-     - Preferred image of |CL| with interactive installer.
+   * - live-desktop.img or live-desktop.iso
+     - Image for live booting desktop version into memory, without requiring installaton.
 
-   * - installer.iso
-     - ISO of |CL| with interactive installer. Only for special cases where ISO image format is required (not for use with a USB key)
-
-   * - live.img
-     - Image for live booting into memory, without requiring installaton.
+   * - live-server.img or live-server.iso
+     - Image for live booting server version into memroy, without requiring installation.
 
 .. list-table:: Table 2: Types of platform-specific |CL| images
    :widths: 15, 85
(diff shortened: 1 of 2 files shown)


source/clear-linux/reference/image-types.rst
```diff
@@ -35,14 +35,11 @@ Table 2 lists the currently available images that are platform specific.
    * - Image Type
      - Description
 
-   * - installer.img
-     - Preferred image of |CL| with interactive installer.
+   * - live-desktop.img or live-desktop.iso
+     - Image for live booting desktop version into memory, without requiring installaton.
```
tpleavitt: The live-desktop will boot into a gnome desktop environment.  The .iso image is read-only and nothing is saved from the session.  If you use the .img to create the USB (using dd or something like that, it creates a RW USB image and changes are saved.  @gtkramer , correct me if that is wrong.

source/clear-linux/reference/image-types.rst
```diff
@@ -35,14 +35,11 @@ Table 2 lists the currently available images that are platform specific.
...
 
-   * - installer.iso
-     - ISO of |CL| with interactive installer. Only for special cases where ISO image format is required (not for use with a USB key)
-
-   * - live.img
-     - Image for live booting into memory, without requiring installaton.
+   * - live-server.img or live-server.iso
+     - Image for live booting server version into memroy, without requiring installation.
```
tpleavitt: the live-server .iso and .img will boot to a command prompt where you will need to log into the system as root the first time.  Same comments apply from above.

Also, you need to change memroy to memory

source/clear-linux/reference/image-types.rst
```diff
@@ -51,31 +48,46 @@ Table 2 lists the currently available images that are platform specific.
...
 
    * - azure-docker.vhd
-     - Virtual Hard Disk for use on Microsoft Azure cloud platform with Docker\* pre-installed
+     - Virtual Hard Disk for use on Microsoft Azure cloud platform with Docker\* pre-installed.
 
    * - azure-machine-learning.vhd
-     - Virtual Hard Disk for use on Microsoft Azure cloud platform with the `machine-learning-basic` bundle installed
+     - Virtual Hard Disk for use on Microsoft Azure cloud platform with the `machine-learning-basic` bundle installed.
 
    * - cloud.img
```
tpleavitt: @gtkramer , Didn't cloud get replaced by cloudguest or do both now exist?

intelkevinputnam: @tpleavitt. As of this release there is still a cloud version. That’s why I left it.

tpleavitt: Let's hear from @gtkramer first.  There is an email thread on dev-public about the images and discusses cloud.img and cloudguest.img.  @bryteise , can you take a look at the proposed change to the image description on the docs site.

intelkevinputnam:
@tpleavitt Removed the cloud image.

source/clear-linux/reference/image-types.rst
gtkramer: That is correct.  We produce the .img for the live-desktop and live-server for this RW capability, but we don't advertise because that is not what marketing wants

source/clear-linux/reference/image-types.rst
gtkramer: @tpleavitt @intelkevinputnam the cloud image is being removed and will be replaced with cloudguest.  We are in a deprecation period with removal scheduled for next Tuesday.  This rename of cloud to cloudguest has already occurred on https://clearlinux.org/downloads.

krisdale:
Resolves issue #532 .

PR #551, opened 2019-06-17
Update KVM download script

tpleavitt:
update the grep string to isolate and download the -kvm.img file


tpleavitt:
This will close #552

Issue #556, opened 2019-06-18
Update assign static IP document to use NetworkManager


puneetse:
**Describe the error/improvement to an existing document or image**
See https://github.com/clearlinux/distribution/issues/915

The current documentation on how to assign a static IP address uses systemd-netword. the default in CL has since switched to NetworkManager as the default NIC manager.

The documentation should be updated to use primarily use NM have  systemd-networkd instructions as optional/reference.

**Additional context**
https://clearlinux.org/documentation/clear-linux/guides/maintenance/assign-static-ip

PR #574, opened 2019-06-27
Update assign static IP doc to include NetworkManager instructions

puneetse:
This PR refreshes the assign static IP address document to include instructions on how to identify network manager program and include instructions for NetworkManager

closes https://github.com/clearlinux/clear-linux-documentation/issues/556


source/clear-linux/guides/maintenance/assign-static-ip.rst
```diff
@@ -1,71 +1,175 @@
...
+   brackets with the appropriate values. *[CONNECTION_NAME]* should be
+   replaced with the NAME from the command above. 
+
+   .. code::
+
+      sudo nmcli connection modify "[CONNECTION_NAME]" \
+      ipv4.method "manual" \
+      ipv4.addresses "[IP_ADDRESS]/[CIDR_NETMASK]" \
+      ipv4.gateway "[GATEWAY_IP_ADDRESS]" \
+      ipv4.dns "[PRIMARY_DNS_IP],[SECONDARY_DNS_IP]"
```
puneetse: @busykai thanks for the review! 

My understanding is that this line won't actually do anything because we're using systemd-resolved for DNS resolution, and we need to provide instructions to configure a [resolved.conf](https://www.freedesktop.org/software/systemd/man/resolved.conf.html). Can you confirm this is true?
```
      ipv4.dns "[PRIMARY_DNS_IP],[SECONDARY_DNS_IP]"
```

busykai: @puneetse, `systemd-resolved` will get the DNS settings from `NetworkManager`. It would be proper to leave these instructions in place: setting static IP means lack of DHCP on the network in some cases and DNS will not be configured.

Alternatively, `systemd-resolved` can be configured to use particular DNS servers, as you mention. However, it's best not to do that and we should not instruct to do so.

Finally, `systemd-resolved` has "fallback" [well-known public] DNS servers (unless they are re-configured/disabled) which is why in absence any DNS configured in either NetworkManager or `systemd-resolved`, hostname resolution will still work.

In case of static IP configuration, it's preferable to configure DNS servers for the interface, because folks who configure static IP should also know the best (closest) DNS service to use for their network.

mvincerx:
LGTM. Thanks, all.

PR #582, opened 2019-07-01
Adds note to clarify BIOS behavior if installing on multiple partitions.

### LINKED ISSUE #543: Add note to baremetal doc. about some BIOS may not show Clear partition if it's installed along-side another OS
Clear allows installation along-side another OS.  During BIOS POST stage, you can select a bootable devices/partitions on a drive.  Some BIOSes can present multiple bootable partitions during this stage, but some will only show 1 bootable partition on a drive.  

For example, on a NUC, if you install Windows and then install Clear along-side, you can press <F10> during POST and it will show Windows Boot Manager and the Clear partition.  You can select Clear to boot it.  If you tried the same method with a Lenovo laptop, it will only present a single boot partition.  So you cannot boot Clear.  It will boot Windows.  I tried this on a couple Lenovos and same result.  

We need to add a `Note` regarding this at this location of the baremetal doc:
`Safe Installation
Use this method to safely install Clear Linux OS on media with available space, or alongside existing partitions, and accept the Default partition schema. If enough free space exists, safe installation is allowed.`

mvincerx:
Closes #543 

### DIFF
--- a/source/clear-linux/get-started/bare-metal-install-desktop/bare-metal-install-desktop.rst
+++ b/source/clear-linux/get-started/bare-metal-install-desktop/bare-metal-install-desktop.rst
@@ -70,8 +70,8 @@ Confirm network connection
 ==========================
 
 Confirm there is a network connection before launching the installer.
-Choose a method: `Wired Connection`, or `WiFi Connected`.
-This guide shows an example of a **Wired Connection**.
+Choose a method: `Wired Connection`, or `WiFi Connected`. This guide shows
+an example of a **Wired Connection**.
 
 #. In the upper right of the top menu bar, select the square icon to view
    Network settings, shown in Figure 2.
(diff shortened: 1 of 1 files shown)


mvincerx:
@bktan8. Can you review this added note per your request from #543 ?

source/clear-linux/get-started/bare-metal-install-desktop/bare-metal-install-desktop.rst
```diff
@@ -243,6 +243,13 @@ Use this method to safely install |CL| on media with available space, or
 alongside existing partitions, and accept the `Default partition schema`_.
 If enough free space exists, safe installation is allowed.
 
+.. note::
+
+   |CL| allows installation along-side another OS. During BIOS POST stage,
+   you can select a bootable device/partitions on a drive. Some BIOSes can
+   present multiple bootable partitions during this stage, but some will
+   show only one bootable partition on a drive. Be aware of limitations.
```
bktan8: This is how I would word it:

|CL| allows installation along-side another OS.  Typically, when you boot your system, you can list and select a bootable device or partition during the BIOS POST stage by pressing an F key.  Some BIOSes may present the |CL| partition and you can select and boot it.  However, there are some BIOSes that may only show the primary partition and you will not be able boot |CL| in this case.  Be aware of this possible limitation with some BIOSes.

PR #598, opened 2019-07-08
Add a note about running stack on other Linux distros

bd-dean:
This PR needs discussion before merging.


source/clear-linux/tutorials/dlrs/dlrs.rst
```diff
@@ -95,11 +95,16 @@ Version compatibility
 
 We validated these steps against the following software package versions:
 
-* |CL| 26240 (Lower version not supported)
+* |CL| 26240 (Minimum supported version)
```
puneetse: Are these components all >= minimum requirements, or very specifically those versions only?
I understand the point of this is the *tested* version, but is the intent that newer version should also work?

bd-dean: Newer versions should work.

bd-dean:
Added section with "uninstallation" instructions.

mvincerx:
LGTM. Thanks, @bd-dean.

PR #606, opened 2019-07-09
Add section on settign a static machine id to telem-guide

### LINKED ISSUE #558: Telemetry page includes a command that no longer works
At the bottom of the page https://clearlinux.org/documentation/clear-linux/guides/telemetrics/telemetry-config#telemetry-config

sudo mkdir -p /etc/telemetrics
sudo echo "unique machine ID" > /etc/telemetrics/opt-in-static-machine-id 
results in a 
-bash: /etc/telemetrics/opt-in-static-machine-id: Permission denied

 - Clear Linux OS version: 30010
 
running 'su -s' and then echo from a # prompt worked.

mvincerx:
- Closes #558.

### DIFF
--- a/source/clear-linux/guides/telemetrics/telem-guide.rst
+++ b/source/clear-linux/guides/telemetrics/telem-guide.rst
@@ -356,7 +356,31 @@ There are three ways to supply the payload to the record.
 
       payload read from stdin
 
+Set a static machine id
+=======================
 
+The machine id reported by the telemetry client is rotated every three days
+for privacy reasons. If you wish to have a static machine id for testing
+purposes, you can opt in by creating a file named "opt-in-static-machine-id"
+in the directory  :file:`/etc/telemetrics/`.
+
+#. Create a directory `telemetrics`.
+
+   .. code-block:: bash
+
+      sudo mkdir -p /etc/telemetrics
+
+
+#. Create the file and replace the "unique machine ID" with your desired
+   static machine ID.
+
+   .. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


source/clear-linux/guides/telemetrics/telem-guide.rst
```diff
@@ -356,7 +356,31 @@ There are three ways to supply the payload to the record.
...
       payload read from stdin
 
+Set a static machine id
+=======================
+
+The machine id reported by the telemetry client is rotated every three days
+for privacy reasons. If you wish to have a static machine id for testing
+purposes, you can opt in by creating a file named "opt-in-static-machine-id"
+in the directory  :file:`/etc/telemetrics/`.
+
```
mvincerx: @bd-dean Would you advise running: `hostnamectl` here beforehand?

bd-dean: I would not recommend :command:`hostnamectl` because it will not report the change to the machine-id, so the user may be worried that telemctl did not pick up their changes.

Systemd and hostnamectl are reading the /etc/machine-id file to pick up & report machine-id. telemctl overrides with the name in the opt-in file but the /etc/machine-id file isn't changed.

mvincerx: Fair enough. Thanks for the detailed response. Any other changes, or are you ready to merge?

bd-dean: Good to go!

PR #613, opened 2019-07-12
Fix error in sequence of figures, live-desktop.

mvincerx:
- repetition of Fig 3. and Fig. 4


mvincerx:
This PR tested the erroneous repetition of currently published Fig. 3 and Fig. 4. However, this PR, if one inspects the file, bare-metal-install-desktop.rst, shows no changes--thus validating that the root cause of this issues is NOT in the  .rst file but rather a web design build error.  Nevertheless I used this opportunity to omit another unused image, `bare-metal-install-beta-29.png`.

Issue #541, opened 2019-06-10
Document iptables firewall and default rulesets


puneetse:
**Do you think Clear Linux documentation needs a new document? Please describe.**
Document which firewall CL uses, no FW rules are set by default, and how to save/restore with systemd units. 

This is not unique to Clear Linux. 

**Should the new document be a guide, a reference, or a tutorial?**
guide

**Describe or provide examples of similar documents, if possible, from other web sites**
https://wiki.archlinux.org/index.php/Simple_stateful_firewall
https://fedoraproject.org/wiki/How_to_edit_iptables_rules
https://help.ubuntu.com/community/IptablesHowTo

**Additional context**
https://community.clearlinux.org/t/firewall-packages/734

mvincerx:
Hold until clarity on new firewall packages. See also FAQ doc for now.

PR #639, opened 2019-07-17
Add a document describing firewall options

puneetse:
This PR adds a document describing firewall options in CL.

Closes https://github.com/clearlinux/clear-linux-documentation/issues/541

### DIFF
--- a/source/guides/network/firewall.rst
+++ b/source/guides/network/firewall.rst
@@ -1,0 +1,201 @@
+.. _firewall:
+
+Firewall
+#########
+
+Firewalls control access to and from systems based on network packet
+attributes like IP address, port, payload and more. 
+
+The `Netfilter <https://netfilter.org/>`_ framework in the Linux kernel
+performs packet filtering and provides the means for implementing a software
+firewall in Linux. |CL-ATTR| has a couple different firewall front-end options
+for managing the Linux firewall.
+
+
+.. contents:: :local:
+    :depth: 2
+   
+
+Default ruleset
+***************
+
+|CL| does not impose a firewall policy out of the box. All traffic is allowed
+inbound and all traffic is allowed outbound. However, `tallow`_ is installed
+by default and may dynamically create a rule temporarily restricting access
+from external hosts.
+
+.. warning::
+
+   Changing firewall configuration can cause abrupt network disconnection. If
+   this happens on a remote host, local recovery may be required.
+   
+   Be sure to test your firewall configuration before committing it
+   permanently to ensure your system will remain accessible remotely, if
+   required.
+
+Firewall software
+*****************
+
+iptables 
+========
+
+:command:`iptables` is a well-known user-space administration tool for
+configuring IPv4 Linux firewall rules. :command:`ip6tables` is the
+complimentary tool for configuring IPv6 Linux firewall rules. 
+
+Below is information on using :command:`iptables` on |CL|: 
+
+#. Make sure the *iptables* bundle is installed
+
+   .. code:: bash
+
+      sudo swupd bundle-add iptables
+
+
+#. Define new iptables rules/chains for the running configuration using the
+   :command:`iptables` command. See :command:`man iptables` for iptables
+   concepts and configuration options.
+
+   Below is a common restrictive firewall configuration which denies all
+   incoming connections, unless the connection was initiated by the host. 
+
+   .. code:: bash
+
+      # Set default chain policies 
+      sudo iptables -P INPUT DROP
+      sudo iptables -P FORWARD DROP
+      sudo iptables -P OUTPUT ACCEPT
+
+      # Accept on localhost loopback device
+      sudo iptables -A INPUT -i lo -j ACCEPT
+      sudo iptables -A OUTPUT -o lo -j ACCEPT
+
+      # Allow established sessions to receive traffic
+      sudo iptables -A INPUT -m conntrack --ctstate ESTABLISHED,RELATED -j ACCEPT
+
+
+#. Test the running firewall configuration to make sure it behaving as
+   you expect. 
+
+#. Run the *iptables-save* service to make the running configuration
+   persistent. This will perform a one-time save of the running configuration
+   to :file:`/etc/iptables.rules` :
+
+   .. code:: bash
+
+      sudo systemctl start iptables-save
+
+#. Enable the *iptables-resolve* service the iptables rules to be
+   automatically applied at boot from the :file:`/etc/iptables.rules` file:
+
+   .. code :: bash
+
+      sudo systemctl enable iptables-restore.service
+
+
+ipset
+=====
+
+`ipset <http://ipset.netfilter.org/>`_ is a framework in the Linux kernel for
+storing and efficiently indexing combinations of IP addresses, networks,
+(TCP/UDP) port numbers, MAC addresses, and interface names. 
+
+IP sets makes writing network policy rules simpler and processing them against
+a large and/or changing sets of hosts more efficient. 
+
+By themselves, IP sets do not enforce network traffic rules but can be used to
+extend iptables rules for matching. It is important to note that the ipset
+must be defined before a netfilter rule can match against it.
+
+* Running IP sets can be manipulated with the :command:`ipset` utility. 
+
+* Custom IP sets can be stored in the :file:`/etc/ipset.conf` file
+
+* IP sets in :file:`/etc/ipset.conf` can be automatically applied at boot by
+  enabling the *ipset* service with the command :command:`sudo systemctl
+  enable ipset`.
+
+See :command:`man ipset` to learn more about using ipsets. 
+
+
+firewalld
+=========
+
+`firewalld <https://firewalld.org/>`_ is based on nftables, the successor to
+iptables and parts of the netfilter framework. The description of firewalld
+helps highlight some of the differences compared to iptables:
+
+    firewalld provides a dynamically managed firewall with support for
+    network/firewall zones to define the trust level of network connections or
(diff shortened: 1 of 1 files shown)


source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+.. contents:: :local:
+    :depth: 2
+   
+
+Default ruleset
+***************
+
+|CL| does not impose a firewall policy out of the box. All traffic is allowed
+inbound and all traffic is allowed outbound. However, `tallow`_ is installed
+by default and may dynamically create rules temporarily restricting SSH
```
ahkok: ..`may dynamically insert one rule temporarily`...

ahkok: `restricting all access` - not just SSH.

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+access.
+
+.. warning::
+
+   Changing firewall configuration can cause abrupt network disconnection, if
+   misconfigured. If this happens on a remote host, local recovery may be
+   required.
+   
+   Be sure to test your firewall configuration before committing it permanently.
```
ahkok: perhaps "and ensure your system will remain accessible remotely in case this is a requirement." ?

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+   :command:`iptables` command. See :command:`man iptables` for iptables
+   concepts and configuration options.
+
+   Below is a common restrictive firewall configuration which denies all
+   incoming connections, unless the connection was initiated by the host. 
+
+   .. code:: bash
+
+      #Delete all existing rules 
+      sudo iptables -F
```
ahkok: I wouldn't do this by default. Definitely note that some programs may insert rules and expect this not to be done randomly.

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+
+      sudo systemctl enable iptables-restore.service
+
+
+ipset
+=====
+
+`ipset <http://ipset.netfilter.org/>`_ is a framework in the Linux kernel for
+storing and efficiently indexing combinations of IP addresses, networks,
+(TCP/UDP) port numbers, MAC addresses, and interface names.
```
ahkok: "By itself, `ipset` does not enforce network traffic rules. It acts as a simple but efficient table that allows `netfilter` rules to match network packets against a large set of entries. This makes writing network policy rules simpler, but also a lot more efficient."

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+
+Custom ipsets do not need to be defined to use iptables but, iptables can be
+extended with IP sets for matching against a large and/or changing sets of
+hosts.
```
ahkok: Note that in order for `netfilter` rules to be inserted, the `ipset` must be created before a `netfilter` rule can match against them. In other words, you need to create the `ipset` rule before you can reference it in a `netfilter` rule.

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+Below is information on using :command:`firewalld` on |CL|: 
+
+#. Make sure the *firewalld* bundle is installed:
+
+   .. code:: bash
+
+      sudo swupd bundle-add firewalld
+
+
+#. Disable *iptables*, *ipset*, and *tallow* services as they conflict with
```
ahkok: Technically not entirely correct. Better would be to modify `tallow.service` to make it *not* create iptables rules - it has an option for that, and have firewalld insert this rule. `man tallow` should document this.

source/clear-linux/guides/network/firewalls.rst
```diff
@@ -0,0 +1,198 @@
...
+When troubleshooting connectivity issues that may be related to firewall
+rules.
+
+* Consider restrictions at the physical network level.
+
+* For inbound connections, make sure your application is listening on the
+  network port you're expecting with :command:`lsof` or :command:`netstat`.
+
+* For outbound connections, make sure the destination host is responding to
+  the network port you're expecting with :command:`telnet`. If the connection
```
ahkok: better: `ping`. Definitely not `telnet`. even `nmap` would be better, I feel.

source/clear-linux/guides/network/firewalls.rst
puneetse: I created https://github.com/clearlinux/tallow/issues/11 to add firewalld support to tallow. 

> and have firewalld insert this rule. man tallow should document this.

I'm uncertain if the `firewalld.ipset` firewalld uses is the same underlying structure as the regular `ipset`. https://firewalld.org/documentation/man-pages/firewalld.ipset.html . So even if the firewalld rule is manually created, I don't think tallow would be updating the same data structure.

```
 $ sudo firewall-cmd --get-ipsets
```

```
$ sudo ipset list
Name: tallow
Type: hash:ip
Revision: 4
Header: family inet hashsize 1024 maxelem 65536 timeout 3600
Size in memory: 88
References: 0
Number of entries: 0
Members:

Name: tallow6
Type: hash:ip
Revision: 4
Header: family inet6 hashsize 1024 maxelem 65536 timeout 3600
Size in memory: 104
References: 0
Number of entries: 0
Members:
```

So my understanding is that in this configuration tallows list wouldn't be used anyways, hence I disabled it. Can you confirm my understanding?

puneetse:
Thanks for the feedback and conversation @ahkok. I've pushed an updated version incorporating the feedback but have an open regard tallow. Could you have a second look?

source/clear-linux/guides/network/firewalls.rst
ahkok: if `ipset` doesn't show firewalld sets then something is really wrong. I don't know what to do in that case - ipsets can't be "temporarily missing" because iptables will not allow you to insert rules referring to it - because it does not exist.

puneetse: So, in its current state doesn't it make sense to disable tallow if firewalld is chosen for use? 
tallow will be creating ipsets/rules that firewalld cannot not use. 

The net result is the same whether tallow is disabled or if is configured not to create rules, right?

ahkok: Disabling tallow never makes sense, imho. We should absolutely drive a solution that doesn't require users to `if (want_feature_x) { disable(feature_y) }`.

If firewalld is stubborn about it, we should patch it to not break tallow (functionally). Especially if firewalld doesn't appear to be able to behave.

Maybe we need to pre-configure firewalld so it runs *after* tallow and somehow happily lives along with tallow.



> The net result is the same whether tallow is disabled or if is configured not to create rules, right?

Not even close? If tallow doesn't make rules, but -say- iptables-restore.service does, tallow is (essentially) fully functional. If you disable tallow, it is entirely non-functional.

puneetse: > If tallow doesn't make rules, but -say- iptables-restore.service does, tallow is (essentially) fully functional.

But the suggestion here is to mask iptables-restore if firewalld is being used so this case wouldn't occur.

In any case, I'll update the document to create a tallow.conf with `nocreate` instead.

mvincerx:
LGTM. @ahkok do you have any other comments--esp on tallow config?

ahkok:
Yes, I don't see or haven't seen anything that shows that `firewalld` will create the needed tallow ipsets.

Passing the `nocreate` flag to tallow will make it omit creating iptables and ipsets. So, you need to at least show how to configure firewalld to create these ipsets and the needed iptables lines, otherwise you should instruct users to disable tallow, because it will just produce errors and do nothing useful.

ahkok:
Note: with `nocreate` tallow will still attempt to insert ipset entries when it needs to block a host temporarily.

puneetse:
I'm looking into the prerequisite issues with firewalld on CL and tallow integration/documentation. Please hold on this PR.

puneetse:
I've rebased this with master and removed the content suggesting to disable tallow. 

Instead, firewalld support has been manufactured for tallow here: https://github.com/clearlinux/tallow/pull/12 and other firewalld bugs resolved here: https://github.com/clearlinux/distribution/issues/1123.

puneetse:
@ahkok can you do another review of this document? I think we can safely publish this now with clearlinux/tallow#12 following separately.

source/guides/network/firewall.rst
```diff
@@ -0,0 +1,204 @@
...
+
+      sudo systemctl mask iptables-restore ipset
+
+
+#. :command:`firewall-cmd` can be used to configure the running or permanent
+   firewall configuration. See the `firewalld documentation
+   <https://firewalld.org/documentation/>`_ to learn more about
+   firewalld concepts and configuration options.
+
+   Below is a common example to allow HTTP traffic in public zones:
```
ahkok: We should default to `HTTPS` here for the example. Not insecure HTTP.

ahkok:
Yeah, I think this is nice and simple now. Let's go and push this.

source/guides/network/firewall.rst
mvincerx: Waiting on final edit. Will merge after your requested change is made.

puneetse: Done.

mvincerx: Cheers.

mvincerx:
LGTM

PR #673, opened 2019-07-29
Update cpu-performance systemctl command

puneetse:
Combine `systemctl enable` and `systemctl start` commands with a single `systemctl enable --now` command.

https://www.freedesktop.org/software/systemd/man/systemctl.html#--now for more information


mvincerx:
LGTM.

PR #717, opened 2019-08-09
Updates to documentation header.

intelkevinputnam:
* Changed Clear Linux Project to all uppercase in substitutions.txt
* Removed link at left of header that went to clearlinux.org
* Added "home" button on right side of header that goes to clearlinux.org


mvincerx:
LGTM. Thanks for the quick turnaround.

PR #718, opened 2019-08-09
Fixes rendering of bullet lists in an admonition (Note).

intelkevinputnam:
Fixed CSS problem where a bullet list inside of an admonition was being overlapped by the admonition text. Thanks @krisdale 

see: https://docs.01.org/clearlinux/latest/guides/maintenance/assign-static-ip.html


mvincerx:
LGTM.

PR #720, opened 2019-08-09
Linkcheck will now only happen on deployment.

intelkevinputnam:
Linkcheck takes a while and only needs to happen on deployment, so added to before_deploy.


mvincerx:
LGTM

PR #734, opened 2019-08-21
Added a magnifying glass icon to the search field and changed placeholder text.

intelkevinputnam:


source/_themes/otc_tcs_sphinx_theme/static/tcs_theme.css
```diff
@@ -284,3 +285,9 @@ div.admonition ul {
 div#trademarks {
     color: gray;
 }
+
+
+#clear-docs-search {
+  background: url(/_static/mag-glass.png) left no-repeat;
```
mvincerx: I think you need to change this url path to current directory, something like:
`background: url(./mag-glass.png) left no-repeat;`

I tested this revision and it worked fine.

intelkevinputnam: Thanks! Fixed.

mvincerx:
LGTM.

Issue #728, opened 2019-08-14
Add "sudo clr-boot-manager update" step prior to list-kernels and set-kernel in Kernel Development doc. 


bktan8:
We need the "clr-boot-manager update" command prior to setting the kernel.  This step will bring the kernel and proper config files over to the EFI partition (/boot), otherwise the files remain in /usr/lib/kernel.  Even though you can list it and set the new kernel, it won't actually show up in the systemd-boot menu and will not boot.  Also, setting a timeout is useful so that you can see the list of selectable kernels.  

So the proper steps are:

```
$ rpm2cpio linux<NAME>-<VERSION>-<RELEASE>.x86_64.rpm | (cd /; sudo cpio -i -d -u -v);
$ sudo clr-boot-manager set-timeout 20
$ sudo clr-boot-manager update
$ sudo clr-boot-manager list-kernels
$ sudo clr-boot-manager set-kernel org.clearlinux.<TARGET>.<VERSION>-<RELEASE>
$ sudo reboot
```

Doc: https://docs.01.org/clearlinux/latest/guides/kernel/kernel-development.html#build-and-install-the-kernel

Section: Build and install the kernel

mvincerx:
@puneetse Can you take this and test it?

PR #740, opened 2019-08-22
Update kernel development document clr-boot-manager commands

puneetse:
This PR updates the clr-boot-manager commands to call update before setting kernel to flush and pending changes to disk and optionally recommends increasing the bootlader timeout for convenience.

Closes https://github.com/clearlinux/clear-linux-documentation/issues/728


mvincerx:
LGTM.

PR #741, opened 2019-08-22
Add make target, venv, for virtualenv; move all dependencies to uppermost requirements.txt

mvincerx:
-  In _scripts/_python for bundle_lister.py
    - Creates global variable PY_VERSION, SHELL default
    - Simplifies Python dev environment by creating venv, novenv targets
- In uppermost directory: 
   - Adds rel path to venv in exclude_patterns in conf.py 
   - Adds virtualenv dependency to uppermost requirements.txt
   - Adds `venv` to upper level .gitignore


mvincerx:
I made some revisions to ensure that the venv is excluded in the make html/htmlall targets. 
This separates the concerns for the Travis job vs. doing local development in a virtualenv.

.travis.yml
```diff
@@ -5,8 +5,9 @@ python:
 
 # command to install dependencies
 install: 
-  - pip3 install -q -r requirements.txt
-  - pip3 install -q -r source/_scripts/_python/requirements.txt
+  - pip3 install -q virtualenv
+  - make venv
+  - source bin/activate/venv
```
mvincerx: Note: Ln 10 command needs to be tested. In case the venv activation doesn't survive into the follow steps, like `scripts`, then we can replace these commands with only: `pip3 install -q -r requirements.txt`, as before.

mvincerx:
Ready. After this is merged, I'll push 1 commit to GH-737, adding a short expl. of using venv.

source/_scripts/_python/Makefile
```diff
@@ -1,9 +1,6 @@
-
 py:
 	python bundle_lister.py
```
intelkevinputnam: Line 2 - Recommend changing this line to: "python3 bundle_lister.py", since it requires python3. That way there's no ambiguity when the it is run, and it isn't dependent on the user's default python.

mvincerx: I'm open to trying this. The top-level Makefile specifies Python 3.6, which references the `py` target. So the py target should execute based on this.

PR #758, opened 2019-08-30
Adds back dlrs.rst in guides/stacks/ to avoid redirect issues.

mvincerx:


mvincerx:
@krisdale. I'm pushing this through. Need it prontissimo!

PR #777, opened 2019-09-10
Updated GA tracking snippet:

krisdale:
- Update to use regex to match hostname for cross domain tracking
- Update to set useAnchor to true (to add session id to anchor instead of query)
- Update to use https when adding analytics.js


krisdale:
Please hold on this for now - confirming one more possible update.

Issue #783, opened 2019-09-12
Improvements for KVM documentation page 


puneetse:
**Describe the error/improvement to an existing document or image**
Some ideas as I was going through https://docs.01.org/clearlinux/latest/get-started/virtual-machine-install/kvm.html to improve:

- Make the list commands to install QEMU/KVM across various distros a table. The bulleted list takes up a lot of room.

- In SSH section, link to https://docs.01.org/clearlinux/latest/guides/network/openssh-server.html and explain why port 10022 is the default for KVM images. 10022 is not the default for all CL systems. 

- Under "Optional: Add the GNOME Display Manager (GDM)" 
  - step 3, use emphasize-lines to better show which lines need to modified. Clean up whitespace after backslash characters which causes issues copy+pasting these sections.  
  - step 4 is vague without giving specific commands.

PR #784, opened 2019-09-12
Update kvm.rst

puneetse:
This PR touches up kvm.rst by:

- Add tables in place of long lists
- Add emphasize-lines in codeblock
- Add commands for refreshing OVMF files

Closes https://github.com/clearlinux/clear-linux-documentation/issues/783

### DIFF
--- a/source/get-started/virtual-machine-install/kvm.rst
+++ b/source/get-started/virtual-machine-install/kvm.rst
@@ -19,34 +19,21 @@ Install QEMU-KVM
 
 #. Log in and open a terminal emulator.
 
-#. Install `QEMU*-KVM` on the host machine. Below are some example distros.
+#. Install `QEMU*-KVM` on the host machine. Below are some examples using
+   different distros:
 
-   * On |CL|:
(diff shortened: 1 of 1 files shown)


source/get-started/virtual-machine-install/kvm.rst
```diff
@@ -102,42 +89,27 @@ Download and launch the virtual machine
```
mvincerx: Starting on Line 66, above, please remove sentences that start with "For example, the |CL| ..." 
I'm sure other developers might appreciate that on the one hand. On the other, I don't want our docs  to support other distro's docs --especially with supporting files. If you want to call out that example, please post those sentences in the Forum. You could also add those lines to an announcement of this much improved `kvm` guide there too.

puneetse: This was the original content and unmodified as part of the changes in this PR. I agree with you and have made the requested changes.

mvincerx: Thank you.

PR #814, opened 2019-09-20
Removes encryption option in notes, screenshots in bare-metal-install-server

mvincerx:


source/get-started/bare-metal-install-server.rst
```diff
@@ -354,14 +354,6 @@ root partition
    If no filesystem exists, the installer will default to `VFAT(FAT32)`
    for `/boot`, and `ext4` for all others.
```
mdhorn: Only the _E needs to be removed, the _F for forcing the formatting is still relevant. 

```suggestion
 .. note::	

      You may also append `_F` to the partition name to force the formatting. i.e.	

      *  `CLR_ROOT_F`: Force the formatting of the root partition prior to use	
      *  `CLR_F_SWAP`: Force the formatting of the swap partition prior to use; helpful when re-using a partition for swap which was previously formatted for a file system.
      *  `CLR_F_MNT_/data`: Force the formatting of the extra data partition prior to use
```

source/get-started/bare-metal-install-server.rst
```diff
@@ -402,7 +394,6 @@ Additional partitions (optional)
       Figure 15: CLR_MNT
 
 #. Alternatively, you may create `CLR_MNT_/srv` or other partitions.
-   Use of the `_E` or `_F` **must** precede `_MNT`.
```
mdhorn: And the `_F` still applies here too
```suggestion
   Use of the `_F` **must** precede `_MNT`.
```

mvincerx: See latest commit -- and my email.

PR #821, opened 2019-09-24
Replace "sudo echo >" with "echo | sudo tee"

claudex:
"sudo echo > " will not work as the current shell will make the redirection and it doen't have the root privileges

### DIFF
--- a/source/guides/network/dpdk.rst
+++ b/source/guides/network/dpdk.rst
@@ -144,14 +144,14 @@ packet buffers.
 
    .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


mvincerx:
LGTM.  I appreciate your contribution and welcome new PRs. 
Going forward, please use descriptive branch names (e.g. claudex-dpdk). We prefer that doc names appear in the branch name.  Here's more [guidance on how to contribute to our docs](https://docs.01.org/clearlinux/latest/reference/collaboration/collaboration.html#contribute-via-github).

Issue #800, opened 2019-09-17
Document about stateless firmware loading


puneetse:
**Do you think Clear Linux documentation needs a new document? Please describe.**
Stateless firmware loading from /etc/firmware is a feature that [unique to clearlinux](https://github.com/clearlinux-pkgs/linux/blob/master/0119-Enable-stateless-firmware-loading.patch)

It would be good to have a document describing its availability and usage.

**Should the new document be a guide, a reference, or a tutorial?**
Recommend a type of document, based on the structure here: Guide

PR #824, opened 2019-09-24
Add document on firmware

puneetse:
Adds a new document on firmware laoding and a reference to it from the stateless document.

Similar documentation for comparison:

- https://wiki.ubuntu.com/Kernel/Firmware
- https://wiki.gentoo.org/wiki/Linux_firmware
- https://wiki.archlinux.org/index.php/Microcode

Closes https://github.com/clearlinux/clear-linux-documentation/issues/800

### DIFF
--- a/source/guides/kernel/firmware.rst
+++ b/source/guides/kernel/firmware.rst
@@ -1,0 +1,96 @@
+.. _firmware:
+
+Firmware
+########
+
+This guide shows how |CL-ATTR| handles firmware and microcode loading.
+
+.. contents::
+   :local:
+   :depth: 1
+
+Overview 
+********
+
+Many devices and system components require firmware or microcode, software
+that runs directly on the device, to function correctly. Because firmware
+loading requires privileged hardware access, the kernel is involved in the
+process.
+
+Firmware does not typically come with source code. Instead, firmware is
+provided as binary blobs which are licensed for free or non-free use.
+
+In |CL| firmware is loaded during device initialization which typically
+happens at boot time. 
+
+
+Included firmware
+*****************
+
+The Linux kernel project contains a repository for firmware binaries that are
+licensed to allow free redistribution. |CL| packages these firmware binaries
+in the `linux-firmware bundles
+<https://clearlinux.org/software?search_api_fulltext=linux-firmware>`_ and
+automatically includes them with the kernel.
+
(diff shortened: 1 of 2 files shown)


source/guides/kernel/firmware.rst
```diff
@@ -0,0 +1,77 @@
...
+
+#. Copy the firmware files including any subdirectories to
+   :file:`/etc/firmware`. It is important to place the firmware files in
+   expected path for proper loading. 
+
+   .. code-block:: bash
+
+      sudo cp -Rv <directory>/<filename>.<fw|bin> /etc/firmware 
+
+
```
bwarden: You might consider an example here, such as something from https://www.intel.com/content/www/us/en/support/articles/000005511/network-and-i-o/wireless-networking.html

bwarden: ...of course pointing out that it's already provided in /usr/lib/firmware by the `linux-firmware-wifi` bundle.

source/guides/kernel/firmware.rst
```diff
@@ -0,0 +1,77 @@
...
+provided as binary blobs which are licensed for free or non-free use.
+
+In |CL| firmware is loaded during device initialization which typically
+happens at boot time. 
+
+
+Included firmware
+*****************
+
+The Linux kernel sources contain firmware binaries that are licensed to allow
```
bwarden: Maybe clarify that it's not the kernel sources containing, but rather the kernel project distributes firmware binaries.

source/guides/kernel/firmware.rst
puneetse: If we're going to add a real example, I rather find an example that isn't an exception/already included and ideally one that has a semi-official source :/.

bwarden: https://storage.microsemi.com/en-us/speed/scsi/linux/aic94xx-seq-30-1_tar_gz.php

bwarden: This is where I'm trying to look:
https://wiki.debian.org/Firmware

But for the most part, it looks like if it's available from an official source, we already have it.

puneetse: I'm going to leave this generic for now. If someone from the community surfaces a good example, we can add it later.

mvincerx:
LGTM

PR #825, opened 2019-09-24
Update K8S tutorial for clarity

### LINKED ISSUE #571: Wordsmithing on the Run Kubernetes* tutorial
The Kubernetes tutorial could use some fleshing out for brand new users. Specific parts include:

Under Configure and run Kubernetes:

Step 2. Enter the commands:
-this could be clearer. What do the commands do, and why run them? Just a simple statement like, 'Enter the commands below to <insert simple explanation here>':

Step 3
-this mentions kubectl, but the user may not know what that is at this point. Including a link or further explanation about how kubectl is a command-line interface used to control Kubernetes would be helpful just so users know what they're doing. It would also be helpful to mention at this step that the kubeadm init command initializes a Kubernetes control-plane node, and how this will be used/useful in using kubernetes. Also, including a step on how to access that control pane would be helpful.

Step 4
-For the uninitiated, what is a RuntimeClass handler?

Under Install pod network add-on:

Under Notes about flannel add-on, it says you must add the following to the kubeadm init command. However, If the user already ran the kubeadm init command as instructed earlier in the tutorial, it is unclear if re-running the command at this point is valid syntax. It would also be helpful if the user knew at this point that the master control pane was setup when they ran the init command, since all they were told was that kubectl was getting configured.

Under Use your cluster:

Some pictures of actually using the cluster, or at least showing the control pane, and where the cluster IP, token, and hash values are displayed, could be helpful to new users. It would also be helpful if there were at least a picture of the on-screen instructions, so that users can be assured that they're looking in the right place.

Under Package configuration customization (optional):

The second paragraph says, " If you edit any of the distribution-provided default files, your changes will be lost in the next system update." I would recommend changing this to reflect the sentiment of the first sentence, so it says something more like, " If you edit any of the distribution-provided default files, your changes will be lost in the next system update unless you move the files to the /etc/ directory."

And finally, just as a general critique on the tutorial, I think it would be helpful to either lengthen this tutorial or create a second one, as mentioned in #272, talking about how to actually use Kubernetes. This tutorial is called 'Run Kubernetes,' and while it does show how to get it up and running, there are a lot of parts and pieces that are never really explored, such as adding pods, deploying containers/services/etc, and several other common Kubernetes tasks that it would be helpful to at least go over briefly.

bd-dean:
Provides more background on commands, corrects errors. Closes #571

### DIFF
--- a/source/tutorials/kubernetes.rst
+++ b/source/tutorials/kubernetes.rst
@@ -3,9 +3,8 @@
 Kubernetes\*
 ############
 
-This tutorial describes how to install, configure, and run the
-`Kubernetes container orchestration system`_ on |CL-ATTR| using CRI+O and
-kata-runtime.
+This tutorial describes how to install, configure, and start the
+`Kubernetes container orchestration system`_ on |CL-ATTR| using CRI+O and kata-runtime.
 
 .. contents::
    :local:
@@ -97,7 +96,7 @@ deployment and your security needs.
       .. note::
 
          Swap will be enabled at next reboot, causing failures in
-         your cluster.
+         your cluster if you choose this method.
 
    or:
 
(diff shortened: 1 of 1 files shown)


mvincerx:
LGTM. Congrats!

PR #833, opened 2019-09-29
bootable-usb.rst small improvement

stokito:
Added a note to keep label as `CLR_ISO` which might save some time for me :)
This is somehow related to https://github.com/clearlinux/clr-installer/pull/554

### DIFF
--- a/source/get-started/bootable-usb.rst
+++ b/source/get-started/bootable-usb.rst
@@ -187,6 +187,8 @@ Burn the |CL| image onto a USB drive
 
 #. Find and select the previously extracted |CL| image file.
 
+#. Check that `Volume label` is set to `CLR_ISO` **Do not change the label as installer relies on it**
+
 #. Click the :guilabel:`START` button. See Figure 2.
 
    .. figure:: /_figures/bootable-usb/bootable-usb-windows-02.png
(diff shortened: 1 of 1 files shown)


mvincerx:
Thanks, @stokito. 

@mdhorn. Will you please confirm USB new min size?

stokito:
BTW, I was able to create and test the USB drive from Ubuntu with UNetbootin and that was much easier than do this with the terminal commands as in the instruction. The only problem was that after uploading the clearlinux iso I needed to add manually the label `CLR_ISO`.
I asked UNetbootin developers to add support of the ClearLinux https://github.com/unetbootin/unetbootin/issues/234 so it will add the label itself.
Anyway, can I add few notes to the instruction that users can use UNetbootin?

mdhorn:
The current desktop ISO is ~2.7GB, but that can change with us enabling offline installations and if DevOp gets requests to add additional bundles. 3GB might be a little tight, but we could probably drop the size to 4GB for the ISO.

stokito:
I changed the note about USB drive size.

mvincerx:
Thanks for the contribution. Going forward, please use a [branch method](https://docs.01.org/clearlinux/latest/reference/collaboration/collaboration.html#contribute-via-github) to submit PRs. It's conventional, and it make testing easier.

PR #841, opened 2019-10-03
Add custom 404 and 403 pages.

krisdale:
- Added static html pages for 404 and 403 server responses.
- Added conf.py setting html_extra_path to copy static _errors folder
  with static 404/403 files into build output


mvincerx:
LGTM. Thanks for the demo in our last meeting.

PR #851, opened 2019-10-08
Revise to allow offline installation in bare-metal-install-desktop and -server

mvincerx:
- -desktop: Removes Fig. 4; renumbers all subsequent Figures
- -desktop: Revises header to Select Additional Bundles and add note on network
- -server: Adds similar preface as above in Select Additional Bundles
- -server: Adds missing `_` in MOUNT point, Fig. 15.


mvincerx:
@mdhorn. Did you have more feedback, other than what we discussed Tues.?  Please LMK.

PR #852, opened 2019-10-09
Adds labels for code blocks used in Clear Linux docs

intelkevinputnam:
image

Code blocks that do not have labels remain the same as before. Currently the  "none" code block is the only type in the docs that does not have a handler. 

Known issues:

1. Code blocks that do not have a label and have line numbers turned on will be mis-aligned. Currently no instances of this.
2. When line numbers are turned on the code block label appears to the right of the line numbers:

image

Current list of code blocks with labels:

* yaml
* bash
* console
* reStructuredText
* Python
* PowerShell
* Diff
* C
* json
* spec
* Perl
* PHP
* xml

### DIFF
--- a/source/_themes/otc_tcs_sphinx_theme/static/tcs_theme.css
+++ b/source/_themes/otc_tcs_sphinx_theme/static/tcs_theme.css
@@ -344,3 +344,67 @@ div#trademarks {
 }
 
 /*End support for code blocks with $ signs that aren't copied with content*/
+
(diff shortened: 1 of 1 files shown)


mvincerx:
LGTM. 
I think this is very helpful--and will be more helpful as we starting integrating more Python, yaml, etc, in code-blocks.  We may need to tweak this in the future because the UI will get busy with features. 
But it's easy to tweak UI stuff later.

krisdale:
LGTM - I'm liking this option a lot!

PR #853, opened 2019-10-09
Remove -version flag from os-install command

bd-dean:
The --version flag points to an old version of Clear Linux, and is not necessary.


mvincerx:
LGTM. Yes, you just covered this in our last meeting. ;)

PR #867, opened 2019-10-15
Add kubectl section for basic use case verification

bd-dean:
This adds a separate section for using kubectl back into the document to support CSP use cases where kubectl is all that is needed.  The section is pulled from a previous release of this doc.


mvincerx:
LGTM.

Issue #538, opened 2019-06-10
How to install and use PHP extensions


puneetse:
**Do you think Clear Linux documentation needs a new document? Please describe.**
This topic has come up a few times so it would be nice to have a document for it. The statelessness of CL is probably the most unique part.

**Should the new document be a guide, a reference, or a tutorial?**
guide

**Additional context**
https://community.clearlinux.org/t/how-to-install-php-extension-ext-imagick/298/

mvincerx:
In progress.

bd-dean:
This issue is satisfied with the comments in the forum https://community.clearlinux.org/t/how-to-install-php-extension-ext-imagick/298/ and doesn't need further documentation here.

puneetse:
I'm reopening this based on this complications described here https://community.clearlinux.org/t/php-fpm-is-it-excluded-from-the-stateless-paradigm-and-not-going-to-be-overwritten-by-swupd/1607/17.

An overall PHP document would be good, clarifying the details here https://github.com/clearlinux/clr-man-pages/blob/master/stateless.7.rst#php-fpm and where PHP values can be defined.

bd-dean:
Closed with #869

PR #869, opened 2019-10-16
Add tutorial for PHP install & config

bd-dean:
@puneetse please take a look, does this need more info on php-fpm? Do I need to add anything for people who have |CL| older than 31020 for forcing PHP to read /etc/php.d paths? 
@MCamp859 does it read ok?  It feels choppy to me, I'd be happy for any suggestions.

### DIFF
--- a/source/tutorials/php.rst
+++ b/source/tutorials/php.rst
@@ -1,0 +1,128 @@
+.. _php:
+
+Installing and Configuring PHP
+******************************
+
+This tutorial describes how to configure and use PHP and PHP extensions on |CL-ATTR|. Although we specifically address PHP this tutorial can serve as a general guide for working with applications in the |CL| :ref:`stateless` environment.
+
+We will focus on enabling PHP and PHP extensions in |CL|.  For a complete :abbr:`LAMP (Linux, Apache, MySQL, PHP)` setup in |CL| refer to the :ref:`web-server-install` section of the :ref:`wordpress` tutorial. For this tutorial you do not need to install MariaDB or WordPress.
+
+.. contents::
+    :local:
+    :depth: 1
+
+
+Prerequisites
+*************
+
+* :ref:`Install <bare-metal-install-desktop>` |CL| on your host system
+* Follow :ref:`web-server-install` to install Apache\* and PHP
+
+
+Configuring PHP
+***************
+
+.. important::
+
+   This tutorial does not cover configuring the php-fpm service. There is a difference between php and php-fpm here. php-fpm sets configuration for the process management and pools of worker bees, and can optionally set some of the values defined in the :file:`php.ini` file. This tutorial is only looking at the PHP configuration directives, not those of php-fpm.
+
+
+By default PHP looks for configuration settings in the :file:`php.ini` file, which resides in the `usr/share/defaults/php/` path. Because |CL| is designed to be a :ref:`stateless` operating system, you must create an optional configuration file to override the default values. Every time :command:`swupd` updates the system it will overwrite changes to the `/usr/share/defaults` file structure.
+
+
+Beginning with |CL| version 31020, PHP has been modified to check for :file:`.ini` files in the /etc/php.d file structure.
+
+You can verify the version of |CL| with :command:`swupd info`
+
+.. code-block:: bash
+
+    sudo swupd info
+
+You will see output similar to:
+
+.. code-block:: console
+
+   Distribution:      Clear Linux OS
+   Installed version: 31310
+   Version URL:       https://cdn.download.clearlinux.org/update
+   Content URL:       https://cdn.download.clearlinux.org/update
+
+
+You can create a :file:`php.ini` as follows:
+
+.. code-block:: bash
+
+   sudo mkdir -p /etc/php.d
+   sudo touch /etc/php.d/php.ini
+
+This file can be edited with any of your specific configuration requirements, and will not be overwritten when swupd performs an update. The `PHP configuration file`_ documentation has complete detail about what you can set in this file.
+
+You can verify the location of the PHP configuration files with the :command:`php --ini` command:
+
+.. code-block:: bash
+
+   php --ini
+
+You should see output like this
+
+.. code-block:: console
+
+   Configuration File (php.ini) Path: /usr/share/defaults/php/
+   Loaded Configuration File:         /usr/share/defaults/php/php.ini
+   Scan for additional .ini files in: /etc/php.d
+   Additional .ini files parsed:      (none)
+
+
+This output indicates that PHP will read the php.ini file from `/usr/share/defaults/php` and will then load any further configuration from :file:`.ini` files in `/etc/php.d/`. We will create a :file:`php.ini` file in `/etc/php.d` for our use, and allow the defaults to be read from `/usr/share/defaults/php/`.
+
+
+Install PHP extensions
+**********************
+
+PHP extensions are compiled libraries designed to enable specific functions in your PHP code. |CL| provides PHP extensions in the :file:`php-extras` bundle.  Install the bundle with swupd:
+
+.. code-block:: bash
+
+   sudo swupd bundle-add php-extras
+
+You can see the list of extensions included in the `php-extras`_ bundle on the |CL| `Store`_.
+
+
+Enable PHP extensions
+*********************
+
+To enable an installed extension we need to add it to the :file:`php.ini` for the composer to use it.
+
+#. Create the :file:`php.ini` file, with the directive to load the php-imagick extension
+
+   .. code-block:: bash
+
+      sudo echo "extension=imagick.so" >> /etc/php.d/php.ini
+
+
+No further detail is required to load the extension, but you must restart the httpd service for PHP to pick up the modification to the `/etc/php.d/php.ini` file.
+
+   .. code-block:: bash
+
+      sudo systemctl restart httpd
+
+You can verify that the imagick extension has been loaded by searching through the runtime list of loaded PHP Modules.
+
+   .. code-block:: bash
+
+      php -m | grep imagick
+
+
+.. note::
+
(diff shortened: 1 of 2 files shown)


source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
+.. _php:
+
+Installing and Configuring PHP
+******************************
```
MCamp859: To align with other tutorials, suggest changing (but note comment below): 
PHP
###

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Prerequisites
+*************
+
+* :ref:`Install <bare-metal-install-desktop>` |CL| on your host system
+* Follow :ref:`web-server-install` to install Apache\* and PHP
+
+
+Configuring PHP
+***************
```
MCamp859: Configure PHP
***************

source/tutorials/php.rst
MCamp859: @mvincerx   Title is very short, but it aligns with the other tutorial titles - is this OK? Maybe "PHP configuration" is better (aligns with Proxy configuration)?

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+
+   This tutorial does not cover configuring the php-fpm service. There is a difference between php and php-fpm here. php-fpm sets configuration for the process management and pools of worker bees, and can optionally set some of the values defined in the :file:`php.ini` file. This tutorial is only looking at the PHP configuration directives, not those of php-fpm.
+
+
+By default PHP looks for configuration settings in the :file:`php.ini` file, which resides in the `usr/share/defaults/php/` path. Because |CL| is designed to be a :ref:`stateless` operating system, you must create an optional configuration file to override the default values. Every time :command:`swupd` updates the system it will overwrite changes to the `/usr/share/defaults` file structure.
+
+
+Beginning with |CL| version 31020, PHP has been modified to check for :file:`.ini` files in the /etc/php.d file structure.
+
+You can verify the version of |CL| with :command:`swupd info`
```
MCamp859: Change "You can verify" to "Verify"  
Suggest you make a similar change to start with an action in the other steps also (Create, Verify, etc).

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+You can verify that the imagick extension has been loaded by searching through the runtime list of loaded PHP Modules.
+
+   .. code-block:: bash
+
+      php -m | grep imagick
+
+
+.. note::
+
+   Enabling an extension only requires that it be installed, added to the php.ini file and that the httpd service is restarted. However extensions may have configuration options.  These will be documented by the extension maintainer.  The options you need can be added to the :file:`/etc/php.d/php.ini` file as described by the documentation for the extension.  Be sure to restart httpd after making changes to the file.
```
MCamp859: Suggest this rewrite:
To enable an extension, you must install it, add it to the php.ini file, and restart the httpd service. However, some extensions may have configuration options, which will be documented by the extension maintainer. Add the options you need to the :file:`/etc/php.d/php.ini` file as described in the extension's documentation. Be sure to restart httpd after changing the file.

source/tutorials/wordpress/web-server-install.rst
```diff
@@ -28,7 +28,7 @@ for more information.
 Install the web-server-basic bundle
```
MCamp859: Change "web-server-basic" to "httpd" to match line 31 below?

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+
+.. important::
+
+   This tutorial does not cover configuring the php-fpm service. There is a difference between php and php-fpm here. php-fpm sets configuration for the process management and pools of worker bees, and can optionally set some of the values defined in the :file:`php.ini` file. This tutorial is only looking at the PHP configuration directives, not those of php-fpm.
+
+
+By default PHP looks for configuration settings in the :file:`php.ini` file, which resides in the `usr/share/defaults/php/` path. Because |CL| is designed to be a :ref:`stateless` operating system, you must create an optional configuration file to override the default values. Every time :command:`swupd` updates the system it will overwrite changes to the `/usr/share/defaults` file structure.
+
+
+Beginning with |CL| version 31020, PHP has been modified to check for :file:`.ini` files in the /etc/php.d file structure.
```
puneetse:
```suggestion

```

I wouldn't bother with this. It's history at this point ;)

source/tutorials/php.rst
puneetse: I would remove the reference to httpd service. That assumes PHP is being used with a web server, and even more specifically apache, which doesn't have to be the case. Maybe Installed, added to php.ini, and reference in the code?

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
+.. _php:
+
+Installing and Configuring PHP
+******************************
+
```
puneetse: Suggest an opening sentence that defined PHP and PHP-FPM with links to their project pages.
PHP is this.
PHP-FPM s that.

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+.. contents::
+    :local:
+    :depth: 1
+
+
+Prerequisites
+*************
+
+* :ref:`Install <bare-metal-install-desktop>` |CL| on your host system
+* Follow :ref:`web-server-install` to install Apache\* and PHP
```
puneetse: A webserver isn't required to use PHP. The prereq here is have the `php-basic` bundle installed.

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Configuring PHP
+***************
+
+.. important::
+
+   This tutorial does not cover configuring the php-fpm service. There is a difference between php and php-fpm here. php-fpm sets configuration for the process management and pools of worker bees, and can optionally set some of the values defined in the :file:`php.ini` file. This tutorial is only looking at the PHP configuration directives, not those of php-fpm.
+
+
+By default PHP looks for configuration settings in the :file:`php.ini` file, which resides in the `usr/share/defaults/php/` path. Because |CL| is designed to be a :ref:`stateless` operating system, you must create an optional configuration file to override the default values. Every time :command:`swupd` updates the system it will overwrite changes to the `/usr/share/defaults` file structure.
```
puneetse:
```suggestion
By default PHP looks for configuration settings in the :file:`php.ini` file, which resides in the :file:`usr/share/defaults/php/` path. Because |CL| is designed to be a :ref:`stateless` operating system, you must create an optional configuration file to override the default values. Every time :command:`swupd` updates the system it will overwrite changes to the :file:`/usr/share/defaults` file structure.
```

`:file:` syntax for paths just to make it render better

source/tutorials/php.rst
```diff
@@ -0,0 +1,128 @@
...
+   Version URL:       https://cdn.download.clearlinux.org/update
+   Content URL:       https://cdn.download.clearlinux.org/update
+
+
+You can create a :file:`php.ini` as follows:
+
+.. code-block:: bash
+
+   sudo mkdir -p /etc/php.d
+   sudo touch /etc/php.d/php.ini
```
puneetse: You might consider having the user-defined php configuration files named something different than `php.ini` just so that it is explicitly clear to the reader that the system php.ini is different from the users. 

It can be confusing for new readers because people tend to refer to any php configuration file as a "php-ini" even if the file isn't named php.ini.

source/tutorials/php.rst
bd-dean: Part of the prerequisites is to install httpd, to constrain the tutorial to a standard environment.  I can add a note after the prereqs to explain that the tutorial assumes httpd, but that the user should make adjustments according to their environment.

source/tutorials/wordpress/web-server-install.rst
bd-dean: thank you!  that saves me a more public d'oh! moment

source/tutorials/php.rst
bd-dean: latest commit changes to "PHP and PHP-FPM"

source/tutorials/php.rst
bd-dean: my bad habit (though maybe not my worst habit)

source/tutorials/php.rst
bd-dean: removed. Always update!

source/tutorials/php.rst
bd-dean: Added in latest commit.  Very brief overview, but helpful context.

source/tutorials/php.rst
bd-dean: I want to leave this in -- people who don't use httpd or other webserver will disregard.  But using httpd gives the opportunity to use a browser to look at phpinfo(), which is useful for people who are following external instructions for modifying php.ini  I added a note in the latest commit indicating httpd is not a requirement.

source/tutorials/php.rst
bd-dean: great re-write, and I've also pulled the httpd reference.

source/tutorials/php.rst
bd-dean: I thought it over during lunch.  You're right. I've removed the httpd references/requirement and added a note with a pointer to the LAMP install tutorial if that's what people are looking for. This is just PHP & PHP-FPM now.

source/tutorials/php.rst
bd-dean: I've chosen the original "my-" prefix, so now the custom ini file is named "my-php.ini" :)

bd-dean:
@mvincerx this should be ready to go, all changes added & it builds on my laptop. Closes #538

mvincerx:
LGTM. Thanks to all for your collaboration on this to drive improvements.

PR #882, opened 2019-10-25
Add Wi-Fi guide

intelkevinputnam:
Shows how to use NetworkManager to set up a Wi-Fi connection using:

* GUI
* TUI
* CLI


source/guides/network/wifi.rst
```diff
@@ -0,0 +1,99 @@
+.. _wi-fi:
+
+Configure Wi-Fi with NetworkManager
```
puneetse: Suggest dropping "with NetworkManager" in the title and using "Wireless network" instead of/in addition to Wi-Fi in the title

source/guides/network/wifi.rst
```diff
@@ -0,0 +1,99 @@
...
+###################################
+
+We recommend using `NetworkManager <https://developer.gnome.org/NetworkManager/stable/NetworkManager.html>`_ to manage network connections. If you choose to
+connect to Wi-Fi while using the live image, your Wi-Fi settings will be
```
puneetse:
```suggestion
connect to Wi-Fi while using the live installer image, your Wi-Fi settings will be
```
Context. A ref link might be good too.

source/guides/network/wifi.rst
```diff
@@ -0,0 +1,99 @@
...
+   :depth: 1
+
+Desktop GUI (Graphical User Interface)
+**************************************
+
+1. Click anywhere on the icons at the right side of the top of the screen to
+   bring up a menu and click on :guilabel:`Wi-Fi Not Connected` then
+   choose :guilabel:`Select Network`.
+
+   .. figure:: /_figures/wifi/wifi-1.1.png
```
puneetse: Suggest cropping this into the upper right corner of the GNOME desktop. The full desktop screenshot is a bit much.

source/guides/network/wifi.rst
```diff
@@ -0,0 +1,99 @@
...
+   .. figure:: /_figures/wifi/nmtui_3.png
+
+#. Select :guilabel:`Back` and hit :kbd:`return`.
+
+   .. figure:: /_figures/wifi/nmtui_4.png
+
+#. Select :guilabel:`Quit` and hit :kbd:`return` to exit. 
+
+   .. figure:: /_figures/wifi/nmtui_5.png
+
```
puneetse: Suggest adding links to https://developer.gnome.org/NetworkManager/stable/nmcli.html and https://developer.gnome.org/NetworkManager/stable/nmcli-examples.html for additional examples.

source/guides/network/wifi.rst
intelkevinputnam: Done.

source/guides/network/wifi.rst
intelkevinputnam: Done.

source/guides/network/wifi.rst
intelkevinputnam: Good idea. Thanks.

intelkevinputnam:
@puneetse I agree about adding wpa_supplicant in a subsequent update. I was able to connect to wifi with wpa_supplicant but ran into problems with being allocated an IP address.

source/guides/network/wifi.rst
```diff
@@ -0,0 +1,99 @@
...
+      IN-USE  SSID                           MODE   CHAN  RATE        SIGNAL  BARS  SECURITY         
+              1st Network                    Infra  1     54 Mbit/s   65      ▂▄▆_  --               
+              2nd Network                    Infra  1     130 Mbit/s  52      ▂▄__  --               
+              3rd Network                    Infra  10    195 Mbit/s  29      ▂___  WPA2             
+
+#. Join the network.
+
+   .. code-block:: bash
+
+      nmcli device wifi connect $SSID password $password
```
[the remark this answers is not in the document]
intelkevinputnam: The problem with this is that it will try to pop up a GUI dialog, which would interfere with terminal only usage. Perhaps a note to use a script or the TUI, if this is a concern.

mvincerx:
LGTM

PR #887, opened 2019-10-30
Fixes a problem where nested collapsible sections do not expand correctly.

intelkevinputnam:
Discovered this in another project. Verified that this solution works to three deep collapsible sections.

Documents impacted:

* Mixer (5 sections in reference section):  https://docs.01.org/clearlinux/latest/guides/clear/mixer.html#references

Changes:

* tcs_theme.css
  * removed transition from CSS as this was blocking the fix from working properly. Removing it also
has no perceivable impact to the expanding and contracting of sections.
* tcs_theme.js (see note below explaining why the whole doc appears to have changed)
  * added function that refreshes the `max-height` attribute of all the collapsible sections from the bottom to the top of the doc (lines 16-27). 
  * added call of function in the button listener that fires when sections are expanded (line 68).

Note: Somehow Windows line endings snuck into tcs_theme.js. Used dos2unix to fix.

### DIFF
--- a/source/_themes/otc_tcs_sphinx_theme/static/tcs_theme.css
+++ b/source/_themes/otc_tcs_sphinx_theme/static/tcs_theme.css
@@ -175,7 +175,7 @@ th,td {
 
 .content-collapse {
   overflow: hidden;
-  transition: max-height 0.2s ease-out;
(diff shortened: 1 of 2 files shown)


mvincerx:
LGTM

Issue #888, opened 2019-10-30
PHP tutorial documentation section is missing nesting or incorrectly nested


JSEHV:
The PHP tutorial documentation on https://docs.01.org/clearlinux/latest/tutorials/index.html is missing nesting or not nested correctly, as the other tutorials. Every chapter is shown as a separate tutorial between OpenVINO and Proxy configuration.

mvincerx:
Good catch, @JSEHV.  @MCamp859.   The PHP tutorial has a few formatting errors.  Will you please apply correct alignment of items below the `.. contents::` directive? This is the reason that the left-side menu does not show correct "nesting," as @JSEHV points out.  Please fix all formatting errors and apply your magic too. Thanks

PR #165, opened 2018-06-16
Create autospec guide and autospec-about to fulfill JIRAs:

mvincerx:
- CLEAR 3076: Autospec concept draft
- CLEAR 3077: Autospec guide

Signed-off-by: Michael Vincerra <michaelx.vincerra@intel.com>

PR #895, opened 2019-11-04
Fixed nesting issue.

MCamp859:
Fixed nesting problem in #888 
Made editorial changes for structure & formatting to align with other tutorials.

Signed-off-by: MCamp859 <maryx.camp@intel.com>

### DIFF
--- a/source/tutorials/php.rst
+++ b/source/tutorials/php.rst
@@ -24,6 +24,30 @@ general guide for working with applications in the |CL| :ref:`stateless`
 environment.
 
 
+Background
+**********
+
+By default, PHP looks for configuration settings in the :file:`php.ini` file,
+which resides in the :file:`usr/share/defaults/php/` path. Because |CL| is a
+:ref:`stateless` operating system, you must create an optional configuration
+file to override the default values. Every time :command:`swupd` updates the
+system, it overwrites changes to the :file:`/usr/share/defaults` file
+structure. To save your configuration options through updates, you must create
+a PHP configuration file in a location that will not be overwritten. The
+recommended location is within the :file:`/etc` file structure, which is why this
+tutorial creates a :file:`/etc/php.d` directory.
+
+The PHP-FPM configuration file is separate from the :file:`php.ini` file used
+by PHP, however it has a similar default path restriction. |CL| installs the
+default :file:`php-fpm.conf` file in :file:`/usr/share/defaults/php`. This
+file with its default values is overwritten during each software update.
+However, PHP-FPM requires that the configuration file exist in that location,
+and, by design, does not read configuration options from a different path.
+This tutorial describes a solution to changing PHP-FPM configuration options
+in |CL|, by manually  overriding the php-fpm.service unit in systemd to pass
+an explicit location to a custom :file:`php-fpm.conf` file.
+
+
 Prerequisites
 *************
 
@@ -159,7 +183,7 @@ below to configure PHP-FPM.
        sudo cp /usr/share/defaults/php/php-fpm.conf /etc/php.d/php-fpm.conf
 
 #.  Make changes to the :file:`php-fpm.conf` file as needed. The
-    `FPM documentation`_ has detail on the configuration options available
+    `FPM documentation`_ has details on the configuration options available
     to PHP-FPM.
 
 #.  Edit the systemd service unit file:
@@ -168,14 +192,10 @@ below to configure PHP-FPM.
 
        sudo systemctl edit --full php-fpm.service
 
-    This opens the php-fpm.service file for systemd in your editor.
+    This opens the :file:`php-fpm.service` file for systemd in your editor.
 
 #.  Change the :command:`ExecStart` configuration to add the
     :command:`--fpm-config` option to point to the custom location:
-
-    .. code-block:: bash
-
-       sudo systemctl edit --full php-fpm.service
 
     .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


source/tutorials/php.rst
```diff
@@ -1,193 +1,251 @@
...
+       Loaded: loaded (/etc/systemd/system/php-fpm.service; enabled; vendor preset: disabled)
+       Active: active (running) since Thu 2019-10-17 13:19:34 PDT; 8min ago
+       Main PID: 14452 (php-fpm)
+       Status: "Processes active: 0, idle: 0, Requests: 0, slow: 0, Traffic: 0req/sec"
+        Tasks: 1
+       Memory: 11.1M
+       CGroup: /system.slice/php-fpm.service
+               └─14452 php-fpm: master process (/etc/php.d/php-fpm.conf)
 
+Background
```
MCamp859: @mvincerx  Moved these 2 paras to streamline the config process and created new section to hold this info because it seemed relevant. Look good to you?

source/tutorials/php.rst
```diff
@@ -1,193 +1,251 @@
...
 
-   .. code-block:: bash
+#.  Change the :command:`ExecStart` configuration to add the
+    :command:`--fpm-config` option to point to the custom location:
 
-      sudo systemctl edit --full php-fpm.service
+    .. code-block:: bash
 
-   .. code-block:: console
+       sudo systemctl edit --full php-fpm.service
```
MCamp859: @bd-dean  This command is the same as the prior step. Can you check your notes for the correct command?

source/tutorials/php.rst
mvincerx: Very nice! LGTM.

mvincerx:
LGTM. As always, I appreciate your judicious edits and rewrites.  
I agree we need a response on Ln 178, as you point out.

source/tutorials/php.rst
mvincerx: @bd-dean Please let use know if this is an error. Thanks.

source/tutorials/php.rst
```diff
@@ -1,193 +1,251 @@
...
 Configure PHP
 *************
 
 .. important::
 
-   This section does not cover configuring the PHP-FPM service. There is a difference between the two configurations. This section is only looking at the PHP configuration directives, not those of PHP-FPM.
+   This section does **not** describe configuration for the PHP-FPM service, which is
+   described later in this guide.
 
+This section creates a :file:`/etc/php.d` directory for all PHP configuration
```
bd-dean: I had more explanation in this paragraph because I want to reinforce how to use stateless.  Shortening it to: "work around the default path restriction" leave people with no context about why CL needs to work around the defaults. I know the concise version reads better but it doesn't explain why we need a PHP tutorial in the first place, which is the intent for this doc.

source/tutorials/php.rst
```diff
@@ -1,193 +1,251 @@
...
 
 Configure PHP-FPM
 *****************
 
-The PHP-FPM configuration file is separate from the :file:`php.ini` file used by PHP. |CL| installs the default :file:`php-fpm.conf` file in :file:`/usr/share/defaults/php` and this file will be overwritten with its default values during each software update. However, PHP-FPM requires that the configuration file exist in that location, and by design will not read configuration options from a different path.
-
-One solution to changing PHP-FPM configuration options in |CL| is to manually override the php-fpm.service unit in systemd to pass an explicit location to a custom :file:`php-fpm.conf` file.
+The PHP-FPM configuration file is separate from the :file:`php.ini` file used
+by PHP, however, it has a similar default path restriction. Follow the steps
+below to configure PHP-FPM.
```
bd-dean: Again, I want to explain --why-- we do things this way, which is not the "standard" linux method for configuring php-fpm. I want more context here, so people don't get frustrated with  the doc just assuming that they have full context on why there's a tutorial for this.

source/tutorials/php.rst
bd-dean: the command is redundant in this place & can be deleted. Or, the block starting at line #165 can be deleted.

source/tutorials/php.rst
```diff
@@ -1,193 +1,251 @@
...
+       Status: "Processes active: 0, idle: 0, Requests: 0, slow: 0, Traffic: 0req/sec"
+        Tasks: 1
+       Memory: 11.1M
+       CGroup: /system.slice/php-fpm.service
+               └─14452 php-fpm: master process (/etc/php.d/php-fpm.conf)
 
+Background
+**********
 
+By default, PHP looks for configuration settings in the :file:`php.ini` file,
```
bd-dean: This should come before the steps explaining how to make the changes. I want to explain why CL has different requirements than other linux distros before explaining how.

source/tutorials/php.rst
mvincerx: OK. Thanks for explaining your approach. Let's go with it.

source/tutorials/php.rst
mvincerx: OK. Thanks for sharing your approach.

source/tutorials/php.rst
mvincerx: What do you think of adding a link to `Background`  in the last line of the [Description](https://github.com/clearlinux/clear-linux-documentation/blob/06e0e44b13105b53239d9ce68fbda845e514e08b/source/tutorials/php.rst#description)?

That could give users who want more explanation background on why we need to create a file in `/etc`.

bd-dean: Adding a link to the Background section to the description is a fine compromise.

source/tutorials/php.rst
mvincerx: Agreed.

MCamp859:
@mvincerx  Moved the Background section upwards and made small text edits. OK to merge now.

mvincerx:
Merging this tutorial today. @bd-dean See note above for follow-up if necessary.

PR #912, opened 2019-11-12
guides/clear/stateless: specify /usr/local as user owned

mrkz:
besides `/var`, `/etc` and `/home`, users are allowed to make
local installations in `/usr/local` without getting swupd
operations to remove such content.

be excplicit about `/usr/local` in the stateless guide.

Relates to https://github.com/clearlinux/distribution/issues/1475

Signed-off-by: Simental Magana, Marcos <marcos.simental.magana@intel.com>

### DIFF
--- a/source/guides/clear/stateless.rst
+++ b/source/guides/clear/stateless.rst
@@ -34,20 +34,22 @@ between user-owned areas and |CL|-owned areas.
 
    Figure 2: With stateless, user and system files are separated on the filesystem.
 
-System areas
-============
-File under the :file:`/usr` directory are managed by |CL| as system files.
+System area
+===========
+Files under the :file:`/usr` directory are managed by |CL| as system files
+(except :file:`/usr/local`).
 Files written under the :file:`/usr` directory by users can get removed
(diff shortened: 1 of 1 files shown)


source/guides/clear/stateless.rst
```diff
@@ -36,18 +36,20 @@ between user-owned areas and |CL|-owned areas.
 
 System areas
 ============
-File under the :file:`/usr` directory are managed by |CL| as system files.
+File under the :file:`/usr` directory are managed by |CL| as system files
```
phmccarty: This was an issue in the previous revision, but the first word of this sentence should be `Files`.

intelkevinputnam:
Thanks for taking this on @mrkz. I recommend a couple of changes for clarity.  

1. Change the section titled "System Areas" to "System Area," since it appears that only /usr is managed by the OS.
2.  The OP specifically called out /opt as an area in question. Even though this is conventionally a user managed area, we should still add to list.

mrkz:
@intelkevinputnam addressed your comments :ok_hand:

PR #914, opened 2019-11-12
Added bash console directive for the clr-boot-manager commands.

bktan8:


bktan8:
Thanks @intelkevinputnam!

PR #927, opened 2019-11-19
Add instructions for virtualenv in readme; revise make venv target in Makefile 

### LINKED ISSUE #905: make venv doesn't work on Clear Linux OS
`make venv` doesn't work on Clear Linux for me because `$(PY_VERSION)` isn't an environment variable out of the box

```
venv:
	virtualenv -p python$(PY_VERSION) venv;
	source venv/bin/activate; \
	pip3 install -r requirements.txt;
```

There are some other ways to determine python version, but I was able to use it without specifying a `-p` parameter at all.

> The default is the interpreter that virtualenv was installed with (/usr/bin/python3)

@mvincerx can you check if this is suitable for your systems too? Thanks for adding this make target. I found it useful.

mvincerx:
- Closes #905

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,6 @@ clean:
 	rm -rf venv
 
 venv:
-	virtualenv -p python$(PY_VERSION) venv;
+	virtualenv venv;\
 	source venv/bin/activate; \
 	pip3 install -r requirements.txt;
-
--- a/README.rst
+++ b/README.rst
@@ -67,6 +67,87 @@ If you want to build the documentation exactly as seen on the website, use
 ``make py`` followed by ``make htmlall``. This builds some
 external dependencies and all supported languages.
 
+Use virtualenv 
+**************
+
+To develop documentation in a ``virtualenv``, use the ``venv`` target.
+The Clear Linux\* OS documentation make target ``venv`` provides a 
+simple development environment that ensures that you have the 
+latest packages and that you manage Python versions separately. Use of the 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -67,6 +67,87 @@ If you want to build the documentation exactly as seen on the website, use
...
+   .. note::
+
+      This assumes Python was already added to your Windows path. 
+
+Install virtualenv 
+------------------
+
+#. Install ``virtualenv``. 
+
+   * On Clear Linux* OS and macOS*:
```
phmccarty: Should this line read `* On |CL| and macOS\*:`, like it was written in the previous section?

Makefile
```diff
@@ -34,7 +34,6 @@ clean:
 	rm -rf venv
 
 venv:
-	virtualenv -p python$(PY_VERSION) venv;
+	virtualenv venv;\
 	source venv/bin/activate; \
-	pip3 install -r requirements.txt;
-
+	pip3 install -r requirements.txt;\
```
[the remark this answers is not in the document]
mvincerx: Sheesh. What was *I* thinking...

mvincerx:
@phmccarty. Thanks for the feedback. Please see latest commit when CI is finished.

README.rst
mvincerx: I realized that all mentions of Clear Linux OS need to be consistently spelled out--because the README is outside scope of  `source` folder .

PR #929, opened 2019-11-20
Add guide for Media Reference Stack

puneetse:
This PR adds a new guide for the Media Reference Stack.

Note: The images render when built locally but not in the GH preview.


mvincerx:
LGTM.

PR #934, opened 2019-11-22
Fixes links that appear to be broken.

intelkevinputnam:
Fixes reference URLS that were being automatically turned into active links that don't go anywhere.

Simple fix - escape the URL by adding "\".


mvincerx:
LGTM

Issue #449, opened 2019-04-08
Wireless connectivity


puneetse:
It would be nice to have a document on the Clear Linux website that shows how to connect to a wireless AP as a client, using wpa_supplicant for example. 

It could highlight and document the current state of network managers in CL and how they may change when using WiFi. (e.g. systemd-networkd vs. NetworkManager)

**Should the new document be a guide, a reference, or a tutorial?**
Guide. 

**Describe or provide examples of similar documents, if possible, from other web sites**
https://wiki.archlinux.org/index.php/WPA_supplicant#Advanced_usage
https://shapeshed.com/linux-wifi/

**Additional context**
Both commandline and GUI examples would be useful.

mvincerx:
@intelkevinputnam Will you please work with @puneetse on this? This is a lower priority, about P3/P4.

intelkevinputnam:
Addressed by PR #882

mvincerx:
@intelkevinputnam will work with @puneetse.

PR #939, opened 2019-12-02
Update to wifi guide to include wpa_supplicant.

intelkevinputnam:

### DIFF
--- a/source/guides/network/wifi.rst
+++ b/source/guides/network/wifi.rst
@@ -217,7 +217,7 @@ or not installed before using wpa_supplicant.
    :file:`/etc/wpa_supplicant/wpa_supplicant-$INTERFACE_NAME.conf`. 
    Use `wpa_passphrase <https://wiki.archlinux.org/index.php/WPA_supplicant#Connecting_with_wpa_passphrase>`_ for a more secure method.
 
-Now, set up ``systemd-networkd.service`` to use this network connection. 
+Now, set up ``systemd-networkd.service`` to request an IP address. 
 
 #. Create the :file:`/etc/systemd/network` directory and
    :file:`/etc/systemd/network/25-wireless.network`. Add the following.
--- a/source/guides/network/wifi.rst
+++ b/source/guides/network/wifi.rst
@@ -8,7 +8,9 @@ choose to connect to Wi-Fi while using the
 :ref:`live installer <bare-metal-install-desktop>` image, your Wi-Fi settings
 will be added to your system during the installation process.
 
-NetworkManager provides three simple methods for configuring Wi-Fi.
+NetworkManager provides three simple methods for configuring Wi-Fi: Desktop,
+CLI, and TUI. :command:`wpa_supplicant` is also available if your prefer a
+more lightweight installation.
 
 .. contents::
    :local:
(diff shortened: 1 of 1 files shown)


intelkevinputnam:
Addresses issue #449

source/guides/network/wifi.rst
```diff
@@ -8,7 +8,9 @@ choose to connect to Wi-Fi while using the
 :ref:`live installer <bare-metal-install-desktop>` image, your Wi-Fi settings
 will be added to your system during the installation process.
 
-NetworkManager provides three simple methods for configuring Wi-Fi.
+NetworkManager provides three simple methods for configuring Wi-Fi: Desktop,
+CLI, and TUI. :command:`wpa_supplicant` is also available if your prefer a
```
puneetse: We should clarify here because NM also uses wpa_supplicant under the hood for wireless connections. 
One idea might be to use headingphrases similar to the static IP doc: https://docs.01.org/clearlinux/latest/guides/network/assign-static-ip.html

- Using NetworkManager
- Using wpa_supplicant + systemd-networkd or Using wpa_supplicant standlone

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
 
+wpa_supplicant
+**************
+
+Make sure NetworkManager is either disabled or not installed before using
+wpa_supplicant.
+
+.. code-block:: bash
+
+   sudo systemctl stop NetworkManager.service
```
puneetse:
```suggestion
   sudo systemctl stop NetworkManager.service
   sudo systemctl mask NetworkManager.service
```

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
 
    .. figure:: /_figures/wifi/nmtui_5.png
 
+wpa_supplicant
+**************
+
+Make sure NetworkManager is either disabled or not installed before using
```
puneetse:
```suggestion
Make sure NetworkManager is either stopped, disabled by masking the service, or not installed
```
Since the service is system distributed and enabled by default, it needs to be masked instead of disabled. 
IIRC uninstalling is difficult.

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
+
+   sudo systemctl stop NetworkManager.service
+
+#. Create a ``wpa_supplicant`` configuration directory.
+
+   .. code-block:: bash
+
+      sudo mkdir /etc/wpa_supplicant
+
+#. Create a minimal :file:`/etc/wpa_supplicant/wpa_supplicant.conf` and add the
```
puneetse: Suggest making this `/etc/wpa_supplicant/wpa_supplicant-$INTERFACE_NAME.conf`. Reasoning in a comment below. This means the step to determine wireless interface needs to come above.

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
+   .. code-block:: bash
+
+      iw dev
+
+#. Complete the configuration process as ``root``.
+
+   .. code-block:: bash
+
+      sudo su
+      wpa_supplicant -B -i $INTERFACE -c /etc/wpa_supplicant/wpa_supplicant.conf
```
puneetse:
```suggestion
      sudo wpa_supplicant -B -i $INTERFACE -c /etc/wpa_supplicant/wpa_supplicant.conf
```

sudo to have access to save configuration under `/etc`

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
+      00:xx:xx:73:7b:45 2412 -59 [WPA2-PSK-CCMP][ESS] Network1
+      00:xx:xx:83:fa:6a 2437 -57 [WPA-EAP-CCMP+TKIP][WPA2-EAP-CCMP+TKIP][ESS]
+      00:xx:xx:83:fa:70 5240 -76 [WPA2-EAP-CCMP][ESS] Network2
+      00:xx:xx:4f:e9:2c 2412 -67 [WPA2-PSK-CCMP][ESS][P2P] Printer
+      00:xx:xx:af:fe:3e 5765 -79 [WPA2-PSK-CCMP][ESS] Network3
+      00:xx:xx:e9:eb:29 2412 -76 [WPA2-PSK-CCMP][ESS] Network4
+      00:xx:xx:26:4a:b9 2412 -79 [WPA2-PSK-CCMP][ESS][P2P] Printer2
+      00:xx:xx:b9:0d:d4 2462 -79 [WPA2-PSK-CCMP][ESS] Network5
+
+#. Set up your network connection.
```
puneetse: Add something like "In this example the network name is Network1 and the password is "Network1Password". 

We also need to call out that this saves the password in plaintext in the `/etc` configuration file. We should point them to `wpa_passphrase`. I'd like to avoid adding another "fork" in the reading flow. I'd like to  deferring to the manpage, but we don't package the man pages for wpa_supplicant.conf. I can look into getting that done, but for now let's at least mention the security consideration.

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
+
+#. Save the configuration, quit :command:`wpa_cli`, and log out of ``root``.
+
+   .. code-block:: bash
+
+      > save_config
+      OK
+      > quit
+      root@clr-live~ # exit
+
```
puneetse: Add a step here to enable the wpa_supplicant.service for future boots

One cool thing about the wpa_supplicant service is that it is a systemd template. So you can "pass" the interface name as the parameter to start a wpa_supplicant service targeting that interface specifically. 

This is the syntax: `systemctl enable --now wpa_supplicant@wlp4s0.service` which expects to read a configuration at `/etc/wpa_supplicant/wpa_supplicant-wlp4s0.conf`

source/guides/network/wifi.rst
```diff
@@ -102,8 +104,119 @@ TUI (Text-based User Interface)
...
+Now, set up ``systemd-networkd.service`` to use this network connection.
```
puneetse:
```suggestion
Now, set up ``systemd-networkd.service`` to use this connection to request an IP address.
```

source/guides/network/wifi.rst
mvincerx: I suggest making explicit what to do with the interface name after run `iw dev`. If the user should see this interface name used after running command on Ln 142, then that should be stated. 
FYI only: In testing, I successfully replaced `$INTERFACE` with the name resulting from `iw dev`.

mvincerx: @puneetse, Are you sure that you want to use the `-` before $INTERFACE_NAME.conf? 
As in: `/etc/wpa_supplicant/wpa_supplicant-$INTERFACE_NAME.conf`

source/guides/network/wifi.rst
intelkevinputnam: Actually you have to be logged in as root - mentioned above.  "sudo" did not work for me.

source/guides/network/wifi.rst
puneetse: Yes, see the reason below. It's a distro-distributed systemd unit template which should be leveraged.

source/guides/network/wifi.rst
puneetse: If you add `ctrl_interface_group=wheel` to the .conf file it should work with sudo. That would be better than having to switch to root.

source/guides/network/wifi.rst
puneetse: Also, we should recommend `chmod 600` on this file.

intelkevinputnam:
Thanks for the feedback @puneetse and @mvincerx . I addressed most of it, but I still need to retest. Will do that Monday.

intelkevinputnam:
Verified current flow on Clear Linux. Good to go from my end.

source/guides/network/wifi.rst
```diff
@@ -102,8 +107,149 @@ TUI (Text-based User Interface)
...
+      00:xx:xx:5d:d9:25 2412 -47 [WPA2-PSK-CCMP][ESS] Network1
+      00:xx:xx:37:25:05 2412 -49 [WPA2-PSK-CCMP][ESS] Network1
+      00:xx:xx:73:7b:45 2412 -59 [WPA2-PSK-CCMP][ESS] Network1
+      00:xx:xx:83:fa:6a 2437 -57 [WPA-EAP-CCMP+TKIP][WPA2-EAP-CCMP+TKIP][ESS]
+      00:xx:xx:83:fa:70 5240 -76 [WPA2-EAP-CCMP][ESS] Network2
+      00:xx:xx:4f:e9:2c 2412 -67 [WPA2-PSK-CCMP][ESS][P2P] Printer
+      00:xx:xx:af:fe:3e 5765 -79 [WPA2-PSK-CCMP][ESS] Network3
+      00:xx:xx:e9:eb:29 2412 -76 [WPA2-PSK-CCMP][ESS] Network4
+      00:xx:xx:26:4a:b9 2412 -79 [WPA2-PSK-CCMP][ESS][P2P] Printer2
+      00:xx:xx:b9:0d:d4 2462 -79 [WPA2-PSK-CCMP][ESS] Network5
```
mvincerx: @puneetse. Can you verify that you tested this section until the end for validity? 
I reached this section without issues, yet after entering `scan` in the `wpa_cli`, the stdout hangs at: 
`<3>CLTRL-EVENT-NETWORK-NOT-FOUND` ... 

Eventually, the initial scan returns `OK`. 
But it still hangs. I don't believe this is a proxy issue. 

I'm alright

intelkevinputnam: @mvincerx I'm pretty sure it doesn't hang, the prompt probably just didn't return right away. Hit return and continue.

mvincerx:
Hey @puneetse.  When can you review like we discussed Wed.?

puneetse:
Sorry for the delay, I do have this on my radar but am prioritizing other work. I'll try to review this weekend/early next week.

intelkevinputnam:
@puneetse thanks for the update. I reviewed and made minor change. Looks good to go.

mvincerx:
@puneetse I like how you simplified steps 5-9.  Please approve or follow-up on comments.

mvincerx:
LGTM

PR #944, opened 2019-12-02
vnc.rst: corrected figure numbering to be sequential

DougTW:
Starting with Figure 5 (was Figure 6) through to Figure 9 (was Figure 10)

Signed-off-by: Doug Martin <doug.martin@intel.com>


mvincerx:
LGTM.

PR #946, opened 2019-12-03
Updated to mount the /boot partition

mns6070:
Added steps to mount and unmount the /boot partition during the repair process.

### DIFF
--- a/source/guides/maintenance/fix-broken-install.rst
+++ b/source/guides/maintenance/fix-broken-install.rst
@@ -43,22 +43,21 @@ Mount root partition, verify, and fix
 
 #. Ensure the system is connected to the network.
 
-#. Mount the system’s root and boot partitions.
+#. Mount the system’s root partition.
 
-   #. To find the root and boot partitions, run:
+   #. To find the root partition, run:
 
       .. code-block:: bash
 
          lsblk -o NAME,LABEL,PARTTYPE,PARTLABEL
 
-      We'll use :file:`/dev/sda3/` as the root partition example and :file:`/dev/sda1/` as the boot partition example.
+      We'll use :file:`/dev/sda3/` as the root partition example.
 
-   #. Next, mount the root partition to the :file:`/mnt` folder and the boot partition to the :file:`/mnt/boot` folder.
+   #. Next, mount the root partition to the :file:`/mnt` folder.
 
       .. code-block:: bash
 
          sudo mount /dev/sda3 /mnt
-         sudo mount /dev/sda1 /mnt/boot
 
 #. Verify that you mounted the correct root partition by checking for some
    files commonly found on |CL| systems.
@@ -76,11 +75,10 @@ Mount root partition, verify, and fix
 
    :ref:`Learn more about how swupd works <swupd-guide>`.
 
-#. After the process is complete, unmount the boot and root partitions:
+#. After the process is complete, unmount the root partition:
 
    .. code-block:: bash
 
-      sudo umount /mnt/boot
       sudo umount /mnt
 
 #. Reboot the system, remove the live desktop USB drive,
--- a/source/guides/maintenance/fix-broken-install.rst
+++ b/source/guides/maintenance/fix-broken-install.rst
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
 
 #. Ensure the system is connected to the network.
 
-#. Mount the system’s root partition.
+#. Mount the system’s root and boot partitions.
 
-   #. To find the root partition, run:
+   #. To find the root and boot partitions, run:
 
       .. code-block:: bash
 
-         lsblk
+         lsblk -o NAME,LABEL,PARTTYPE,PARTLABEL
 
-      We'll use :file:`/dev/sda3/` as the root partition example.
+      We'll use :file:`/dev/sda3/` as the root partition example and :file:`/dev/sda1/` as the boot partition example.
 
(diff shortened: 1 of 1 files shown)


mvincerx:
@mns6070 Thanks for continuing to review our docs and contribute.
@bktan8. This seems like a logical approach. Could you please review this as you're the original author?

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
 
 #. Ensure the system is connected to the network.
 
-#. Mount the system’s root partition.
+#. Mount the system’s root and boot partitions.
```
bktan8: swupd does not manage the EFI partition.  So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
...
 
-   #. To find the root partition, run:
+   #. To find the root and boot partitions, run:
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
...
 
       .. code-block:: bash
 
-         lsblk
+         lsblk -o NAME,LABEL,PARTTYPE,PARTLABEL
```
bktan8: Good suggestion.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
...
 
-      We'll use :file:`/dev/sda3/` as the root partition example.
+      We'll use :file:`/dev/sda3/` as the root partition example and :file:`/dev/sda1/` as the boot partition example.
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
...
 
-   #. Next, mount the partition to the :file:`/mnt` folder.
+   #. Next, mount the root partition to the :file:`/mnt` folder and the boot partition to the :file:`/mnt/boot` folder.
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -43,21 +43,22 @@ Mount root partition, verify, and fix
...
 
       .. code-block:: bash
 
          sudo mount /dev/sda3 /mnt
+         sudo mount /dev/sda1 /mnt/boot
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -75,10 +76,11 @@ Mount root partition, verify, and fix
 
    :ref:`Learn more about how swupd works <swupd-guide>`.
 
-#. After the process is complete, unmount the root partition:
+#. After the process is complete, unmount the boot and root partitions:
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount the boot partition.

source/guides/maintenance/fix-broken-install.rst
```diff
@@ -75,10 +76,11 @@ Mount root partition, verify, and fix
...
 
    .. code-block:: bash
 
+      sudo umount /mnt/boot
```
bktan8: swupd does not manage the EFI partition. So we it's not necessary to mount and unmount the boot partition.

source/guides/maintenance/fix-broken-install.rst
mvincerx: I agree. This is helpful in viewing the output. In summary, please delete all other changes on a new commit to this PR, and then we'll accept for merge. Cheers, @mns6070 .

mns6070:
What happens if there is a kernel update from a newer release level?  Doesn't swupd try to update the kernel by calling clr-boot-manager, which does manage the EFI boot partition?

bktan8:
Correct.  clr-boot-manager manages the EFI partition.

mns6070:
> Correct. clr-boot-manager manages the EFI partition.

Just want to make sure I understand the workflow completely.  If someone runs the swupd repair command referenced here without /boot EFI parition mounted and there is an update to the kernel, swupd updates the kernel and calls clr-boot-manager to finish the kernel update.  At that point if /boot is not mounted, clr-boot-manager writes the changes to /boot, which would be on the root partition and not on the EFI partition because Clear Linux is booted from the recovery media and not from the local disk, correct?

bktan8:
"swupd repair" doesn't do anything with kernel updates.  It simply goes through /usr and /var and verify all of the files against the manifest of the version of Clear installed and make sure they are correct.  If they are not, swupd will repair them.

mns6070:
Okay.  Just making sure.

mns6070:
Everything should be fixed now.

mvincerx:
LGTM. Thanks, all.

PR #956, opened 2019-12-10
Windows command sha512 -> SHA512

extremepayne:
The command fails (on my computer) unless the hash algorithm is passed as uppercase.
As per https://answers.microsoft.com/en-us/windows/forum/all/certutil-hashfile-sha1-failed/1a9f31fa-bdf5-4c63-8e72-91608606ba14
Also, I didn't try to modify line 176 and I don't think I did, but for some reason, it shows up on the diff.

### DIFF
--- a/source/guides/maintenance/download-verify-decompress.rst
+++ b/source/guides/maintenance/download-verify-decompress.rst
@@ -142,7 +142,7 @@ checksum file designated with the suffix `-SHA512SUMS`.
 
     .. code-block:: bash
 
-        CertUtil -hashfile ./clear-[version number]-[image type].[compression type] sha512
+        CertUtil -hashfile ./clear-[version number]-[image type].[compression type] SHA512
 
 #.  Manually compare the output with the original checksum value shown in
     the downloaded checksum file and make sure they match.


extremepayne:
I have no idea why the build failed; is that CI non-essential? Is there a way to allow failures?

mvincerx:
I appreciate you reviewing our docs. 
In testing this on my Windows 10 Enterprise, it made no difference whether or not `sha512` was lower case or upper case. Both worked fine. I'm running `OS Version: 10.0.17763 N/A Build 17763`. 

If you don't mind, open cmd prompt and type: `systeminfo`. Then check which  Windows build you're using. If you're using an older build, by chance, feel free to research potential incompatibilities, and let us know what you discover. I'll leave this open until 12-16, Monday.

extremepayne:
Thank you so much for following up on this. Yes, I'm on win7. But that's the reason why I want to upgrade. EOL is looming, and I want Linux. Is there a specific reason to keep it lowercase? Otherwise, I'd recommend changing it to the most compatible version.
I'm running 7 Home Premium 6.1.7 SP1 build 7601, with 214 hotfixes (I won't list out the hashes for brevity).

mvincerx:
We can go with uppercase. I notice that Windows Powershell docs use uppercase for verifying SHA sums. Also, please push again to this PR. The failure wasn't related to your PR; different random error.

extremepayne:
Alright, done. Thanks again for your help; random errors can be confusing. You can rebase away that extra commit if you want.

mvincerx:
LGTM

PR #961, opened 2019-12-17
cpu-performance.rst: add thermal configuration section

guowangy:
add thermal configuration section for cpu-performance page

### DIFF
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -205,23 +205,11 @@ technology, and convert to the thermal_conf.xml configuration format used
 by thermald. It's a closed-source project, and unable to be packaged as bundle
 in Clear Linux OS, so we need to follow below steps to generate configuration.
 
-Intel DPTF requires BIOS support, it's typically used by laptop.
+Intel DPTF requires BIOS support, it's typically used by laptops.
 The first step is to make sure your machine's BIOS has DPTF feature
 and is enabled.
 
-Then, check thermald version is equal to or above version 1.9.1:
-
-.. code:: bash
-
-   thermald --version
-
-If its version is not up-to-date, please upgrade to latest Clear Linux version.
-
-.. code:: bash
-
-   sudo swupd update
-
-Next step, generate thermal configuration as below:
+Then generate thermal configuration as below:
 
 .. code:: bash
 
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -196,32 +196,34 @@ group:
 Thermal configuration
 =====================
 
-Better thermal control and performance can be archived by providing platform
-specified configuration to thermald.
+Better thermal control and performance can be achieved by providing platform
+specific configuration to thermald.
 
 `Linux DPTF Extract Utility`_ is a companion tool to thermald, This tool can
-make use of "Intel ® Dynamic Platform and Thermal Framework (`Intel DPTF`_)"
+make use of :abbr:`Intel® Dynamic Platform and Thermal Framework (`Intel DPTF)`
 technology, and convert to the thermal_conf.xml configuration format used
 by thermald.
 
-First, make sure thermald version is equal or above 1.9.1:
+First, make sure thermald version is equal to or above version 1.9.1:
 
 .. code:: bash
 
    thermald --version
 
-Otherwise, you need to add an option :command:`--ignore-default-control` to
+If its version is not updated, append the  :command:`--ignore-default-control` to ``ExecStart``,  as shown below.
 thermald:
 
 .. code:: bash
 
    # edit thermald.servcie
-   vim /usr/lib/systemd/system/thermald.service
+sudo  vim /usr/lib/systemd/system/thermald.service
    # append --ignore-default-control option to ExecStart line, like below
-   # ExecStart=/usr/bin/thermald --no-daemon --dbus-enable --ignore-default-control
+.. code:: bash
+   ExecStart=/usr/bin/thermald --no-daemon --dbus-enable --ignore-default-control
 
    # reload
-   sudo systemctl daemon-reload
+ .. code-block:: bash
+    sudo systemctl daemon-reload
 
 Then, generate thermal configuration as below:
 
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -221,6 +221,7 @@ Next step, generate thermal configuration as below:
 
 .. code:: bash
 
+   sudo swupd bundle-add acpica-unix2  # install acpi tools
    git clone https://github.com/intel/dptfxtract.git
    cd dptfxtract
    sudo acpidump > acpi.out
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -216,7 +216,8 @@ Then, check thermald version is equal to or above version 1.9.1:
    thermald --version
 
 If its version is not up-to-date, please upgrade to latest Clear Linux version.
-.. code-block:: bash 
+
+.. code:: bash
 
    sudo swupd update
 
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -216,6 +216,9 @@ Then, check thermald version is equal to or above version 1.9.1:
    thermald --version
 
 If its version is not up-to-date, please upgrade to latest Clear Linux version.
+.. code-block:: bash 
+
+   sudo swupd update
 
 Next step, generate thermal configuration as below:
 
--- a/source/guides/maintenance/cpu-performance.rst
+++ b/source/guides/maintenance/cpu-performance.rst
@@ -193,8 +193,8 @@ group:
    After adding a new group, you must log out and log back in for the new group
    to take effect.
 
-Thermal configuration
-=====================
+Enhanced thermal configuration
+===============================
 
 Better thermal control and performance can be achieved by providing platform
 specific configuration to thermald.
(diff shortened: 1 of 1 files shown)


source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
    After adding a new group, you must log out and log back in for the new group
    to take effect.
 
+Thermal configuration
+=====================
+
+Better thermal control and performance can be archived by providing platform
```
mvincerx:
```suggestion
Better thermal control and performance can be achieved by providing platform
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+specified configuration to thermald.
+
+`Linux DPTF Extract Utility`_ is a companion tool to thermald, This tool can
+make use of "Intel ® Dynamic Platform and Thermal Framework (`Intel DPTF`_)"
```
mvincerx:
```suggestion
make use of :abbr:`Intel® Dynamic Platform and Thermal Framework (`Intel DPTF)`
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+technology, and convert to the thermal_conf.xml configuration format used
+by thermald.
+
+First, make sure thermald version is equal or above 1.9.1:
```
mvincerx:
```suggestion
First, make sure thermald version is equal to or above version 1.9.1:
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+
+.. code:: bash
+
+   thermald --version
+
+Otherwise, you need to add an option :command:`--ignore-default-control` to
```
mvincerx:
```suggestion
If its version is not updated, append the  :command:`--ignore-default-control` to ``ExecStart``,  as shown below.
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+thermald:
+
+.. code:: bash
+
+   # edit thermald.servcie
+   vim /usr/lib/systemd/system/thermald.service
```
mvincerx:
```suggestion
sudo  vim /usr/lib/systemd/system/thermald.service
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+   # append --ignore-default-control option to ExecStart line, like below
```
mvincerx: Remove Ln 220 (it now appears above).

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+   # ExecStart=/usr/bin/thermald --no-daemon --dbus-enable --ignore-default-control
```
mvincerx: Make Ln 221 a separate example, like so: 
```suggestion
.. code:: bash
   ExecStart=/usr/bin/thermald --no-daemon --dbus-enable --ignore-default-control
```
Assure that this line is **NOT** commented out with a '#'

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+
+   # reload
+   sudo systemctl daemon-reload
```
mvincerx:
```suggestion
 .. code-block:: bash
    sudo systemctl daemon-reload
```

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+
+Then, generate thermal configuration as below:
+
+.. code:: bash
+
+   git clone https://github.com/intel/dptfxtract.git
+   cd dptfxtract
+   sudo acpidump > acpi.out
+   acpixtract -a acpi.out
+   sudo ./dptfxtract *.dat
```
mvincerx: When I tested these commands, `sudo ./dptfxtract *.dat` produced the following: 
```
...
No valid tables found
```
Is this an error?

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
...
+
+thermald configuration files will be generated and saved to
+:command:`/etc/thermal/` folder. Restart thermald service to take effect.
+
+.. code:: bash
+
+   sudo systemctl restart thermald.service
+
```
mvincerx: After running this command, I ran again the command ``thermald --version``. The output was ```1.9``, not ``1.9.1``. 

Should ``thermald`` have been upgraded? Or, was this prevented due to the error on "tables" from above?   
Can you include a command that will confirm that the above changes took effect? In other words: What change should be noticeable from the standpoint of a user?

mvincerx:
@puneetse. Can you please test this too? This should also support your other new doc. 😉

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
    After adding a new group, you must log out and log back in for the new group
    to take effect.
 
+Thermal configuration
+=====================
+
+Better thermal control and performance can be archived by providing platform
+specified configuration to thermald.
```
puneetse:
```suggestion
specific configuration to thermald.
```

source/guides/maintenance/cpu-performance.rst
puneetse: I see this was not packaged natively in Clear Linux because it is closed-source: https://github.com/clearlinux/distribution/issues/1073 

We should mention why it is not bundled and the limitations.

source/guides/maintenance/cpu-performance.rst
puneetse: I'd rather make sure the version in Clear Linux is at least this version before publishing to avoid cluttering with static version check information. That sort of thing tends to be irrelevant quickly with the pace that Clear Linux updates move.

source/guides/maintenance/cpu-performance.rst
puneetse: See comment above about avoiding this whole section, but in Clear Linux editing file under `/usr/` is not safe and prone to being overwritten by system update. See the documentation on stateless. 

If you have to do this in the future on Clear Linux system, a better approach is to create a modified service file in `/etc/`. The `systemctl edit --full thermald.service` can help do this neatly. Even better would be a drop-in directory for the unit. https://www.freedesktop.org/software/systemd/man/systemd.unit.html

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,53 @@ group:
    After adding a new group, you must log out and log back in for the new group
    to take effect.
 
+Thermal configuration
```
puneetse: Maybe a better title for this section is "Enhanced thermal configuration"

source/guides/maintenance/cpu-performance.rst
guowangy: DPTF need BIOS support. Most Intel laptop support this feature.
Typical path is: Configuration -> DPTF (Enabled)

source/guides/maintenance/cpu-performance.rst
guowangy: Current version in Clear Linux is 1.9.
I submit a patch to upgrade to 1.9.1, and is under review now.

source/guides/maintenance/cpu-performance.rst
guowangy: No, modifying `thermald.service` will not change `thermald` version, it's a workaround for version under `1.9`.

User can check `thermald` status by `systemctl status thermald.service`, and find log like:
```
Dec 06 05:44:54 icl-clr-debug-i5 thermald[459]: [WARN]Using generated /etc/thermald/thermal-conf.xml.auto
```
it indicates that the configure file has been used by `thermald`.

guowangy:
Workaround of old thermald version is removed since I create a version update patch to thermal-daemon package, once it merged, workaround is not needed.

source/guides/maintenance/cpu-performance.rst
mvincerx: That's fair.  Please include a statement to that effect. Something along the lines of "This is a workaround..."

source/guides/maintenance/cpu-performance.rst
guowangy: I can not find such a guide/article to enable DPTF in BIOS. And menu entry might changed from different vendor.

And from the project of dptf for chromimum OS, here is what they says:

> Requirement: DPTF requires corresponding ACPI support in BIOS. Not all
Intel based platforms support DPTF in BIOS. Please contact your BIOS vendor
to see if DPTF is enabled in your system.

https://github.com/intel/dptf

source/guides/maintenance/cpu-performance.rst
guowangy: I just remove this workaround part, and suggest user to upgrade to latest Clear Linux version.

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,59 @@ group:
...
+
+If its version is not up-to-date, please upgrade to latest Clear Linux version.
+
+Next step, generate thermal configuration as below:
+
+.. code:: bash
+
+   git clone https://github.com/intel/dptfxtract.git
+   cd dptfxtract
+   sudo acpidump > acpi.out
```
puneetse: Can you include a step to ensure the bundle that contains these tools (acpica-unix2) is installed (i.e. `sudo swupd bundle-add acpica-unix2`

guowangy: fixed in [0abdc54](https://github.com/clearlinux/clear-linux-documentation/pull/961/commits/0abdc549c8f9f41c68bca48e6b623365ea6604d2)

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,60 @@ group:
...
+and is enabled.
+
+Then, check thermald version is equal to or above version 1.9.1:
+
+.. code:: bash
+
+   thermald --version
+
+If its version is not up-to-date, please upgrade to latest Clear Linux version.
+
```
mvincerx:
```suggestion
.. code-block:: bash 

   sudo swupd update

```

source/guides/maintenance/cpu-performance.rst
mvincerx: OK. Please add the command to update, shown on Line 219 above.

source/guides/maintenance/cpu-performance.rst
```diff
@@ -193,6 +193,60 @@ group:
...
+.. code:: bash
+
+   sudo systemctl restart thermald.service
+
+check whether the configuration is in used.
+
+.. code:: bash
+
+   sudo systemctl status thermald.service
+
```
mvincerx: Please state what message (stdout) the user should see if configuration is in use. (This can be console output too.)

guowangy: already mentioned in L244.

mvincerx:
LGTM. 
@puneetse, please review when time permits.

mvincerx:
@puneetse, Are you satisfied with the latest commits in response to your feedback?

source/guides/maintenance/cpu-performance.rst
mvincerx: Given that the option to enable may vary per the BIOS, I'll let this stand as is. Refinements to this will be addressed in a follow-up GH Issue.

mvincerx:
LGTM.

PR #962, opened 2019-12-17
Update for DLRS V5 release

bd-dean:
Adds updated information for the DLRS V5 Release, and updates instructions for using Selden and OpenVINO to serve models.


mvincerx:
LGTM

PR #965, opened 2019-12-20
Update readme URL and remove redundant Seldon section

bd-dean:
URL points to incorrect github repo for stacks readme.

### DIFF
--- a/source/guides/stacks/dlrs.rst
+++ b/source/guides/stacks/dlrs.rst
@@ -378,95 +378,6 @@ We provide `DLRS PytorchJob`_ examples that use the Deep Learning Reference Stac
 
 
 
-Using Kubeflow Seldon and OpenVINO* with the Deep Learning Reference Stack
-**************************************************************************
-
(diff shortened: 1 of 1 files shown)


source/guides/stacks/dlrs.rst
```diff
@@ -998,7 +909,7 @@ Related topics
 
 .. _DLRS Terms of Use: https://clearlinux.org/stacks/deep-learning/terms-of-use
 
-.. _DLRS Release notes: https://github.com/clearlinux/dockerfiles/blob/master/stacks/dlrs/releasenote.md
+.. _DLRS Release notes: https://github.com/intel/stacks/tree/master/dlrs
```
mvincerx: Hey @bd-dean. Just checking. Do I understand correctly that the purpose is to link to release notes only on GitHub? 
I believe this is so. If that's so, then do you want to link directly to the release notes? Like so: 
https://github.com/intel/stacks/blob/master/dlrs/clearlinux/releasenote.md

PR #986, opened 2020-01-10
Fixes link that was hard coded to point to clearlinux.org/documentation ...

intelkevinputnam:
Not sure why this broke now. Could be a failure to forward or a page that was finally removed on the clearlinux.org/documentation side.

Now uses :ref: to same page.

Also made some opportunistic formatting changes.


mvincerx:
LGTM. Nice catches.

PR #1002, opened 2020-01-22
/etc/sysctl.d/60-k8s.conf -> /usr/lib/sysctl.d/60-k8s.conf

cornerot:
the directory /etc/sysctl.d does not exist

### DIFF
--- a/source/tutorials/kubernetes.rst
+++ b/source/tutorials/kubernetes.rst
@@ -66,7 +66,7 @@ deployment and your security needs.
 
 #. Enable IP forwarding to avoid kubeadm `preflight check`_ errors:
 
-   Create (or edit if it exists) the file :file:`/etc/sysctl.d/60-k8s.conf`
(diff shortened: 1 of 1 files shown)


mvincerx:
Thanks for submitting a PR.  Clear Linux OS ships with a mostly empty `/etc` directory. This [stateless design](https://docs.01.org/clearlinux/latest/guides/clear/stateless.html#stateless) is what differentiates ours from other Linux distros--and keeps the OS lean.  Part of this approach means that when a user runs `sudo swupd update`,  some `/usr` directories will be overwritten.  Rather, we  recommend adding custom config to the `/etc/` 
[explained here](https://docs.01.org/clearlinux/latest/guides/clear/stateless.html#default-configurations). Our docs won't accept your PR because it assumes an approach that's different from our all of our documents, and it doesn't follow the Clear Linux development philosophy. That said, you're welcome to configure your OS however you like. We welcome other contributions to our docs.  Closing this PR.

PR #1013, opened 2020-01-31
Updated time.rst

eadamsintel:
When I tried this I had an error when I tried to set ntp to true that polkit was missing. Adding that bundle allowed the step to complete.

### DIFF
--- a/source/guides/network/time.rst
+++ b/source/guides/network/time.rst
@@ -13,7 +13,7 @@ the default :abbr:`NTP (Network Time Protocol)` servers cannot be reached.
 
    .. code-block:: bash
 
-      sudo swupd bundle-add sysadmin-basic
+      sudo swupd bundle-add sysadmin-basic polkit
 
 #. Set your time zone. This example uses Los Angeles.
 
--- a/source/guides/network/time.rst
+++ b/source/guides/network/time.rst
@@ -13,7 +13,7 @@ the default :abbr:`NTP (Network Time Protocol)` servers cannot be reached.
 
    .. code-block:: bash
 
-      sudo swupd bundle-add sysadmin-basic polkit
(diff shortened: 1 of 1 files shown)


puneetse:
Did you run the systemctl command with `sudo` when you got the polkit error? If I remember correctly, the polkit message is a symptom of invoking without having sudo permission first.

I'd be inclined not to document it this way because I don't think polkit is actually a timedatectl/ntp dependency.

eadamsintel:
Good point. Using `polkit` is probably not the best way to do this.  The instructions don't indicate to use sudo so this is what happened.

```
$ timedatectl set-ntp true
Failed to set ntp: Unit polkit.service not found.
$ sudo timedatectl set-ntp true
```

Behind the corporate proxy I had issues getting the time to sync even once.

puneetse:
Ah, so we should add`sudo` to the `timedatectl` command in the document and add direct to add polkit I think.

eadamsintel:
I am ok with that. I did a pull request from the web interface so I am not sure the easiest way to modify it is.

puneetse:
You can go back to the file in your branch : https://github.com/clearlinux/clear-linux-documentation/blob/eadamsintel-patch-1/source/guides/network/time.rst

and just hit the pencil icon to edit again

mvincerx:
In case you're also using the CLI, don't forget to run `git pull [--rebase] upstream master` after.

mvincerx:
👍 to @puneetse 's comment above re `sudo` and no need for `polkit` in this case. When I removed polkit, CL still allowed me to set the timezone. @eadamsintel LMK if you need any help on the PR.

mvincerx:
LGTM. Thanks, @eadamsintel.

PR #1018, opened 2020-02-04
Fix Windows path in vmw player doc

bktan8:
Missing backslashes in the Use |CL| pre-built VMware image tab.


bktan8:
Bad PR.  Closing and restarting.

PR #1024, opened 2020-02-06
update to current settings

krz:
The current CUDA Toolkit is v10.2 and now supports gcc8. I was able to install it only when the --samples option was removed. This is advice from https://community.clearlinux.org/t/cuda-toolkit-installation-error/1699

### DIFF
--- a/source/tutorials/nvidia-cuda.rst
+++ b/source/tutorials/nvidia-cuda.rst
@@ -70,7 +70,7 @@ Refer to the `NVIDIA documentation on CUDA system requirements
 <https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#system-requirements>`_
 for the latest kernel and compiler compatibility. 
 
-For example, CUDA 10.1 on a system with the latest Linux kernel requires GCC7,
+For example, CUDA 10.2 on a system with the latest Linux kernel requires GCC8,
 which is older than the default GCC version for |CL|.
 
 Install the compatible version of GCC, if required:
@@ -79,7 +79,7 @@ Install the compatible version of GCC, if required:
 
    .. code:: bash
 
-      sudo swupd bundle-add c-extras-gcc7
+      sudo swupd bundle-add c-extras-gcc8
       
 #. Create the directory :file:`/usr/local/cuda/bin`:
 
@@ -95,8 +95,8 @@ Install the compatible version of GCC, if required:
 
    .. code:: bash
    
-      sudo ln -s /usr/bin/gcc7 /usr/local/cuda/bin/gcc
-      sudo ln -s /usr/bin/g++7 /usr/local/cuda/bin/g++  
+      sudo ln -s /usr/bin/gcc8 /usr/local/cuda/bin/gcc
+      sudo ln -s /usr/bin/g++8 /usr/local/cuda/bin/g++  
 
 
 Downloading and Installation
@@ -173,7 +173,6 @@ installed.
       
       sudo ./cuda_<VERSION>_linux.run \
       --toolkit \
-      --samples \
       --installpath=/opt/cuda \
       --no-man-page \
       --override \


mvincerx:
@puneetse I saw that you were responding to this issue on the forum. Please LMK your thoughts here.

puneetse:
I've seen this reported a few times now so the change makes sense. Line 236 in the document should also be removed since it mentions samples.

krz:
It is also important to mention that the NVIDIA driver needs to be compiled without the **--no-nvidia-modprobe** argument to get CUDA functionality. This should be either mentioned here or on the NVIDIA driver docs page https://docs.01.org/clearlinux/latest/tutorials/nvidia.html
Honestly, this argument should be completely removed. It took me quite a while to figure this out.

source/tutorials/nvidia-cuda.rst
```diff
@@ -79,7 +79,7 @@ Install the compatible version of GCC, if required:
 
    .. code:: bash
 
-      sudo swupd bundle-add c-extras-gcc7
+      sudo swupd bundle-add c-extras-gcc8
```
mvincerx: 👍

mvincerx:
> It is also important to mention that the NVIDIA driver needs to be compiled without the **--no-nvidia-modprobe** argument to get CUDA functionality. This should be either mentioned here or on the NVIDIA driver docs page https://docs.01.org/clearlinux/latest/tutorials/nvidia.html
> Honestly, this argument should be completely removed. It took me quite a while to figure this out.

@krz Regarding your first sentence, the flag **--no-nvidia-modprobe** *is* documented at this link under the step [Run the installer with the advanced options below](https://docs.01.org/clearlinux/latest/tutorials/nvidia.html#install-the-nvidia-drivers). Your third sentence is not clear to me because it appears to contradict the first. Please clarify.

puneetse:
On the NVIDIA driver documentation, the `--no-nvidia-modprobe` switch is specifically included because the nvidia-modprobe installation path conflicts with the Clear Linux system locations. It can easily get removed after installation. And as far as I could tell, nvidia-modprobe is an outlier in that it doesn't obey the other switches from nvidia-installer to allow putting in an alternate location under `/opt/` or `/usr/local` where it won't get removed (i.e. even when `--utility-prefix=/opt/nvidia` is set `nvidia-modprobe` gets installed under `/usr`) If you find an update or some installation switch that changes this behavior, please let us know.

nvidia-modprobe exists to create the character devices if they're missing (e.g. headless system). This can be done without a specific utility but CUDA looks for nvidia-modprobe even if it is not required (i.e. the character device is already created). So the best compromise we could think of was documenting how to manually build nvidia-modprobe and install it under `/usr/local`. Which is in the nvidia-cuda doc. 

@lebensterben  may have additional thoughts.

krz:
I'm not an expert on what **--no-nvidia-modprobe** exactly does but I can tell you that removing it (and thus activating modprobe I guess) results in no problems using the current 440.59 NVIDIA drivers.

What I want to say is that most new Clear Linux users probably follow the NVIDIA drivers docs page (https://docs.01.org/clearlinux/latest/tutorials/nvidia.html) on how to install the proprietary drivers.
However, if they follow these steps they will not be able to install the CUDA functionality later, because the NVIDIA drivers guide suggests to install with **--no-nvidia-modprobe**. 
So if a user follows the NVIDIA driver guide first and then later the CUDA guide, this user will get an non-functioning CUDA system because nowhere it is mentioned that modprobe is required at compilation.
This is what happened to me.

lebensterben:
>I'm not an expert on what **--no-nvidia-modprobe** exactly does but I can tell you that removing it (and thus activating modprobe I guess) results in no problems using the current 440.59 NVIDIA drivers. ...

@krz You're exactly right. This flag won't affect the NVIDIA driver but affect CUDA functionality.

I found this problem first when I try to run CUDA examples. Many examples need to detect your current GPU model, which relies on modprobe binary.

And I agree with @puneetse that the best compromise is to how to preserve nvidia-modprobe without contaminating the /usr/bin directory.

The solution could be as simple as removing --no-modprobe and move /usr/bin/modprobe elsewhere. I suggest to have the nvidia-modprobe moved to /opt/nvidia/bin because that's where other nvidia binaries are installed.

mvincerx:
Will merge. @puneetse please open separate issue related  to note on `modprobe`.

Issue #918, opened 2019-11-14
Direct users to use LTS kernel in NVIDIA driver tutorial


lebensterben:
Currently the documentation **mentioned** that LTS kernel is more likely to be compatible with the NVIDIA proprietary driver comparing to the frequently updated native kernel.

It'd be better if it would be rephrased to a tone that more strongly **suggests** and **directs** the user to use LTS kernel than just **mentioning** this fact.

For users, they're less likely to see graphical desktop environment failed to start because of some conflicts between the kernel and NVIDIA proprietary driver.

puneetse:
In my experience the opposite can also be true pretty often: pairing against LTS kernels can result in breakages that have long since been fixed in newer version pairings. 

That is why I avoided being more direct with the language.

mvincerx:
Hey @puneetse. Please let me know if you plan to address this or not. Thanks
https://docs.01.org/clearlinux/latest/tutorials/nvidia.html#install-dkms

puneetse:
I'm not going to address this right now. We'll keep our ears open to datapoints coming in. If it starts to sway one way or the other, we can certainly revisit the issue.

PR #1028, opened 2020-02-13
Added recommendation to run LTS kernel for proprietary NVIDIA drivers…

eeeple:
… (see https://github.com/clearlinux/distribution/issues/1725)

### DIFF
--- a/source/tutorials/nvidia.rst
+++ b/source/tutorials/nvidia.rst
@@ -57,6 +57,10 @@ are known to be problematic.
    documentation. As a simple workaround, some systems can disable one of the
    graphics devices or NVIDIA Optimus* in the system firmware.
 
+NVIDIA is known for not keeping up with the kernel release cycle. As such, it 
+is recommended to run the :ref:`LTS kernel <compatible-kernels>` (*kernel-lts*) 
+instead of the native kernel (*kernel-native*) (see `this GitHub 
+issue <https://github.com/clearlinux/distribution/issues/1725>`_).
 
 **See the** `Troubleshooting`_ **section for more known issues and solutions.**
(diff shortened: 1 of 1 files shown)


puneetse:
See also https://github.com/clearlinux/clear-linux-documentation/issues/918

I'm hesitant to add this as a blanket recommendation because these tend to be point-in-time issues and on the other hand I have seen long-since solved issues trying to use LTS kernels. So, I think the recommendation should stay loose or at most recommend trying a *different* kernel variant. 

There is this sentence further below in the document that gets inherited from the DKMS section: `The Long Term Support (LTS) kernel variant is more likely to remain compatible between updates with NVIDIA drivers.` which could be moved up into the Known issues section instead.

eeeple:
@puneetse Thanks for the input ! I've removed the previous paragraph and rephrased with something more in line with the remark in the dkms section.

mvincerx:
I'm fine with the latest push. @eeeple Replace this PR with a branch, instead of submitting a PR on your `master` branch. After @puneetse approves, [create a branch as on a standard Git workflow](https://docs.01.org/clearlinux/latest/collaboration/collaboration.html#contribute-via-github), and submit a new PR with the approved change,

mvincerx:
@eeeple. I'd like to give you credit for this, but you need to submit a PR with a new branch in order for this to be accepted. Please indicate if you want to do so by Thurs. 02-20-2020. Otherwise, I'll close this PR and we'll make a new one with your edit.

eeeple:
@mvincerx I'm sorry, didn't have the time to push before. Submitted as #1034 from new branch. Thank you 😄

PR #1043, opened 2020-02-24
GitHub Actions Support

intelkevinputnam:
This is the first step toward migrating all of Clear Linux Docs CI/CD to GitHub actions.


mvincerx:
Looks great. Let's test these around 13:30 today. 

My only feedback so far is whether we should integrate a release tag --for automation-- in the Chinese docs --or for the English version.  There could be one for Language Versions or for Document Versions. At this point, either of these seem like a TBD.

We could create a separate GH Action for a release tag to be published. Thoughts?

mvincerx:
LGTM

Issue #1026, opened 2020-02-07
Kata is no longer the default runtime


puneetse:
**Describe the error/improvement to an existing document or image**

The default has changed to have Kata no longer be the default runtime for Docker. 
https://github.com/clearlinux/distribution/issues/1451

The Kata documentation (https://docs.01.org/clearlinux/latest/tutorials/kata.html) needs to be updated to reflect that and how to opt-in to using kata.

PR #1054, opened 2020-02-27
Update kata doc after default runtime change

puneetse:
This PR updates the kata document to have instructions for opting-in to use kata since it is no longer the default runtime.

Closes https://github.com/clearlinux/clear-linux-documentation/issues/1026

### DIFF
--- a/source/tutorials/kata.rst
+++ b/source/tutorials/kata.rst
@@ -27,7 +27,7 @@ the :ref:`bare metal installation tutorial<bare-metal-install-desktop>`.
 If you have Clear Containers installed on your |CL| system, then follow the
 :ref:`migrate Clear Containers to Kata Containers tutorial<kata_migration>`.
 
-Before you install any new packages, update |CL| with the following command:
+Update |CL| with the following command:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


source/tutorials/kata.rst
```diff
@@ -37,89 +37,90 @@ Install Kata Containers
...
-   If you use a proxy server and your proxy environment variables are already
-   set, run the following commands as a shell script to configure Docker:
+#. Create a systemd drop-in configuration for Docker:
 
    .. code-block:: bash
 
-      docker_service_dir="/etc/systemd/system/docker.service.d/"
-      sudo mkdir -p "$docker_service_dir"
-      cat <<EOF | sudo tee "$docker_service_dir/proxy.conf"
+      sudo mkdir -p /etc/systemd/system/docker.service.d/
```
eadamsintel: We might want to consider doing this step using /etc/docker/daemon.json to set the default runtime.  It is a tough call because we generally use systemd units but `kata-deploy` uses `/etc/docker/daemon.json` to add its runtimes.   I don't think you have to do a daemon-reload when using dockers daemon.json.

How to do this is at https://docs.docker.com/engine/reference/commandline/dockerd/

```
	"no-new-privileges": false,
	"default-runtime": "runc",
	"oom-score-adjust": -500,
	"node-generic-resources": ["NVIDIA-GPU=UUID1", "NVIDIA-GPU=UUID2"],
	"runtimes": {
		"cc-runtime": {
			"path": "/usr/bin/cc-runtime"
		},
		"custom": {
			"path": "/usr/local/bin/my-runc-replacement",
			"runtimeArgs": [
				"--debug"
			]
		}
	},
	"default-address-pools":[
		{"base":"172.80.0.0/16","size":24},
		{"base":"172.90.0.0/16","size":24}
	]
}
```

source/tutorials/kata.rst
```diff
@@ -37,89 +37,90 @@ Install Kata Containers
...
+         Default Runtime: kata-runtime
 
-You do not need to manually configure the runtime for Docker, because
-it automatically uses the runtime supported by the system.
 
 Troubleshooting
 ===============
 
+- If you are behind a HTTP proxy server, in a corporate setting for
+  example, please refer to the `Docker proxy instructions`_ .
```
eadamsintel: I had the period render on a separate line for some reason.

source/tutorials/kata.rst
puneetse: That's a really good point. I will add a note about the `daemon.json` method but leave the CLI method as the primary in the steps.

source/tutorials/kata.rst
```diff
@@ -37,89 +37,96 @@ Install Kata Containers
...
 
-   If you use a proxy server and your proxy environment variables are already
-   set, run the following commands as a shell script to configure Docker:
+   .. note:: 
+
+      The method below uses a systemd drop-in configuration to add a
+      command-line (CLI) parameter to the Docker daemon for setting the
+      default runtime. Alternatively, the default runtime can be set in the
+      `Docker daemon configuration file
+      <https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file>`_.
```
eadamsintel: I am good with this but if you wanted to push one more commit it would be helpful to mention that if settings are set in two different places that conflict then Docker daemon will fail to start.

eadamsintel: Otherwise I approve it as it is.

puneetse: Good point. I've added a sentence calling it out.

puneetse:
@mvincerx this is ready

mvincerx:
LGTM.

PR #1060, opened 2020-03-05
Correct filenames

lavacano:
its the small things that matter, 2 missing hyphens; fixing them turned my symlinks from red to green.  I only hope these are the correct names!


mvincerx:
Thanks for the sharp eyes on this, @lavacano. Keep coming back to the CL Docs. 😄 
@puneetse, thanks for providing the reference link.

PR #1063, opened 2020-03-06
Added another step

lavacano:
run ❯ nvidia-settings after installing the app profiles to /etc/nvidia you'll get this:

(nvidia-settings:16608): GLib-GObject-CRITICAL **: 01:26:53.561: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

ERROR: nvidia-settings could not find the registry key file. This file should have been installed along with this driver at
       either /usr/share/nvidia/nvidia-application-profiles-440.59-key-documentation or
       /usr/share/nvidia/nvidia-application-profiles-key-documentation. The application profiles will continue to work, but
       values cannot be prepopulated or validated, and will not be listed in the help text. Please see the README for possible
       values and descriptions.

which is a problem https://devtalk.nvidia.com/default/topic/1072456/linux/all-gui-apps-low-fps-when-mouse-is-inactive-x-org-1-20-7-rtx2080ti/

I'm not sure why the author chose to install them in /etc/nvidia they aren't looked for there.  I'm not sure how much buggery it would be to keep /usr/share sacred but this one link is the difference between gui apps running like dirt.

### DIFF
--- a/source/tutorials/nvidia.rst
+++ b/source/tutorials/nvidia.rst
@@ -286,6 +286,11 @@ Install the NVIDIA drivers
 
       ln -sv /opt/nvidia/share/applications/nvidia-settings.desktop $HOME/.local/share
 
+#. Optional: Link NVIDIA application profiles to where they are searched for
+
+   .. code-block:: bash
+
+      ln -s /etc/nvidia /usr/share
 
 Updating
 ********


puneetse:
-1 from me.

The path for this keys-doc file specifically is hardcoded in `nvidia-settings`: [source](https://github.com/NVIDIA/nvidia-settings/blob/376bb212229f3801fe343632e15f12ba9fe209bc/src/gtk%2B-2.x/ctkappprofile.c#L3669) and doesn't follow the other nvidia-installer switches. 

I don't think this is the right recommendation for CL to give users. Files put under `/usr/share` can get clobbered if a `swupd repair` is run with the `--picky` flag. The error is also non-blocking behavior. From a tutorial perspective, it can be a step/note to be done post-install by users so they are aware of the implication. 

@lebensterben and I had a conversation in IRC last night about this too. He purposed a method of maintaining a patched version of `nvidia-settings` for the scripts he maintains.

lebensterben:
@lavacano yep I think adding a symbolic link to /usr/share is not hygienic.

And suppose the user invoked `swupd repair --picky`, `nvidia-settings` would suddenly stop working correctly.

It's better to avoid causing those potential troubles by maintaining a patched version of `nvidia-settings`, which would be simple as replacing one line.

lavacano:
I agree the patch isnt perfect but neither will be maintaining a patchset for this driver, it iterates about ever 3 months.  

I do not think this is limited to just nvidia-settings if you read through the devtalk forum post the symptoms are an irresponsible desktop when composition pipeline is disabled (for example watching a simple gif will get 1 fps)  I think the drivers also hunt for these files :(

lavacano:
Perhaps it's as easy as installing the profiles to /etc/nvidia/nvidia-application-profiles-rc.d 

https://http.download.nvidia.com/XFree86/Linux-x86/325.15/README/profiles.html

lavacano:
@lebensteben @puneetse nvidia-bug-report.sh has the driver hunt logic:

```
cat /usr/local/bin/nvidia-bug-report.sh | grep application
        append_file_or_dir_silent "$DIR/.nv/nvidia-application-profiles-rc"
        append_file_or_dir_silent "$DIR/.nv/nvidia-application-profiles-rc.backup"
        append_file_or_dir_silent "$DIR/.nv/nvidia-application-profiles-rc.d"
        append_file_or_dir_silent "$DIR/.nv/nvidia-application-profiles-rc.d.backup"
        append_silent "$DIR/.nv/nvidia-application-profile-globals-rc"
        append_silent "$DIR/.nv/nvidia-application-profile-globals-rc.backup"
append_file_or_dir_silent "/etc/nvidia/nvidia-application-profiles-rc"
append_file_or_dir_silent "/etc/nvidia/nvidia-application-profiles-rc.d"
append_file_or_dir_silent /usr/share/nvidia/nvidia-application-profiles-*-rc
```

all we need to do on the install step is mkdir -p /etc/nvidia/nvidia-application-profiles-rc and modify the installer 1 liner to be `--application-profile-path=/etc/nvidia/nvidia-application-profiles-rc` 

This can be improved with some version control and the nvidia-settings ONLY looks for the profiles in /usr/share so it is still a problem for the app but with this simple modification the driver level should be fixed I will close this and the other push request and put in another patch later today unless you guys beat me to it.

puneetse:
@lavacano the tutorial already does that. The `application-profile-path` doesn't address the `nvidia-application-profiles-key-documentation` which is what the error is complaining about.

lavacano:
> @lavacano the tutorial already does that. The `application-profile-path` doesn't address the _key_ file (`nvidia-application-profiles-key-documentation`) which is what the error is complaining about.

nvidia-settings complains it wants them in /usr/share/nvidia.  Not the case with the driver

The driver finds either /etc/nvidia/nvidia-application-profiles-rc (not /etc/nvidia/nvidia-application-profiles-440.64-key-documentation or /etc/nvidia/nvidia-application-profiles-440.64-rc) which is where your version of the documentation installs them.  I've pushed a patch that installs them to a path that the driver *does* look for them. 

See https://github.com/clearlinux/clear-linux-documentation/pull/1064

lebensterben:
> Oh that's right, that keys path is hard coded: https://github.com/NVIDIA/nvidia-settings/blob/376bb212229f3801fe343632e15f12ba9fe209bc/src/gtk%2B-2.x/ctkappprofile.c#L3669
Not a blocking error obviously.

@puneetse found this last night

lavacano:
@lebensterben @puneetse https://github.com/NVIDIA/nvidia-settings/issues/50

PR #1064, opened 2020-03-06
Makes driver find application profiles

lavacano:
muchas importante

@puneetse @lebensterben

### DIFF
--- a/source/tutorials/nvidia.rst
+++ b/source/tutorials/nvidia.rst
@@ -225,7 +225,7 @@ server must be configured to use the content under
               ModulePath      "/opt/nvidia/lib64/xorg/modules"
       EndSection
       EOF
-      
+
 Install the NVIDIA drivers
 ==========================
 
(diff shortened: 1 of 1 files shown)


source/tutorials/nvidia.rst
```diff
@@ -225,7 +225,12 @@ server must be configured to use the content under
               ModulePath      "/opt/nvidia/lib64/xorg/modules"
       EndSection
       EOF
+      
+#. Make a place for application profiles to live (previously installed files need to be removed before updating drivers)
 
+   .. code-block:: bash
+   
+      sudo mkdir -p /etc/nvidia/nvidia-application-profiles-rc.d
```
puneetse: Can you test if the nvidia-installer will make this path for you when it doesn't exist? If it doesn't, I'd prefer omitting this step and letting the installer do its thing.

lavacano: I agree. give me 20.

mvincerx:
LGTM. Thanks, all.

PR #1066, opened 2020-03-09
Update instructions for X11 forwarding and SFTP

### LINKED ISSUE #1027: OpenSSH X11 instructions need to be upated
The settings in the OpenSSH documentation to enable X11 forwarding do not include all the settings to disable the security that Clear Linux defaults to.

The details are in the `man stateless` page. 

**Additional context**
https://community.clearlinux.org/t/no-attempt-to-forward-x-over-ssh-has-succeeded/2938

bd-dean:
closes #1027

### DIFF
--- a/source/guides/network/openssh-server.rst
+++ b/source/guides/network/openssh-server.rst
@@ -3,7 +3,7 @@
 Enable and configure SSH service
 ################################
 
-This guide describes how to set up SSH service.
+This guide describes how to set up the SSH service.
 
 .. contents::
    :local:
@@ -14,7 +14,7 @@ Overview
 
 The :command:`openssh-server` bundle provides the OpenSSH package that
 enables an SSH service in |CL-ATTR|. Remote users require an SSH service to be
-able to use an encrypted login shell.
+able to use an encrypted login shell. The SSH daemon has all of its configuration built in and no template configuration file is present on the file system.
 
 |CL| enables the `sshd.socket` unit, which listens on port 22 by default
 and starts the OpenSSH service as required. The first time OpenSSH starts, it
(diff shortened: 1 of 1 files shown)


source/guides/network/openssh-server.rst
```diff
@@ -104,32 +104,20 @@ Enable SFTP
...
 
-#. Create a systemd drop-in directory for the SSHD service:
-
-   .. code-block:: bash
-
-      sudo mkdir -p /etc/systemd/system/sshd@.service.d
-
-#. Create the following file:
-   :file:`/etc/systemd/system/sshd@.service.d/sftp.conf`
+#. Create the following file, if it does not already exist:
```
eadamsintel: I think this step 1 should match the step 1 of the root login and x11-forwarding sections. You are doing the exact same thing so I think the wording should match exactly.  I really doubt anyone has to `mkdir /etc/ssh` as described in the root login section and that command would fail anyways without sudo. I recommend changing the root login and x11-forwarding section step 1/2 to match your step 1 in the sftp section.

source/guides/network/openssh-server.rst
```diff
@@ -104,32 +104,20 @@ Enable SFTP
...
+      subsystem sftp /usr/libexec/sftp-server
 
-#. Reload systemd configuration:
 
-   .. code-block:: bash
-
-      sudo systemctl daemon-reload
-
-Congratulations! The SFTP subsystem is enabled.
+Congratulations! The SFTP subsystem is enabled. You do not need to restart the sshd service.
```
eadamsintel: I would add this note to the root login section as well.

bd-dean: Good call -- adds balance.  Will add with next commit.

source/guides/network/openssh-server.rst
bd-dean: Agree -- fixed in next commit.

mvincerx:
LGTM. Let's review today. ;)

mvincerx:
LGTM

PR #1068, opened 2020-03-10
nvidia.rst: Add a section to restore ld.so.conf

lebensterben:
During installation, `ld.so.conf` was modified to include `/etc/ld.so.conf.d`, under which
a file `nvidia.conf` was created so that `ld` looks for shared library installed by NVIDIA driver.

This commit adds information on removing `nvidia.conf` file.
Users could optionally remove the line `include /etc/ld.so.conf.d/*.conf` that we added to
`/etc/ld.so.conf`.


mvincerx:
LGTM.

PR #1069, opened 2020-03-10
nvidia.rst: Correct the path for desktop files

lebensterben:
Desktop files should be created under `$HOME/.local/share/applications` instead of `$HOME/.local/share`,
at least on my system only the former one works.

This commit fixed the path for both installation and uninstallation.


mvincerx:
LGTM.

PR #1070, opened 2020-03-10
nvidia.rst: Remove `chmod +x` on installer

lebensterben:
With `sudo sh NVIDIA-Linux-x86_64-<VERSION>.run`, we don't need to change its permission.

This commit makes it a bit more convenient for users.

### DIFF
--- a/source/tutorials/nvidia.rst
+++ b/source/tutorials/nvidia.rst
@@ -240,17 +240,12 @@ Install the NVIDIA drivers
 
       cd ~/Downloads/
 
-#. Make the :file:`NVIDIA-Linux-x86_64-<VERSION>.run` file executable.
-
-   .. code-block:: bash
-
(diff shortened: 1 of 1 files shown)


puneetse:
@bwarden do you have any thoughts on the differences between these methods? Is there any risk to be aware of invoking a root shell?

bwarden:
> 
> 
> @bwarden do you have any thoughts on the differences between these methods? Is there any risk to be aware of invoking a root shell?

From a security standpoint, I don't think it's an additional risk. There is the possibility that they add bash-specific commands to it, or change the installer script to a different language, or even a compiled binary. All of those, while unlikely, would break this.

puneetse:
Thanks for the perspective. I think this is a reasonable. If the compatibility of the installer script changes in the future, we can update.

mvincerx:
LGTM.

PR #1078, opened 2020-03-17
nvidia.rst: Add a section on GCC compatibility

lebensterben:
Address the issue mentioned in https://github.com/clearlinux/distribution/issues/1835

This commit helps the user to detect a potential problem early, and provides a workaround.

### DIFF
--- a/source/tutorials/nvidia.rst
+++ b/source/tutorials/nvidia.rst
@@ -67,6 +67,36 @@ are known to be problematic.
 Installation
 ************
 
+Verify GCC Compatibility
+========================
+
+The installation requires building a kernel module with GCC, and furthermore
+you should compile the kernel module with the same GCC version that is used
+to compile the kernel.
+
+To get the current GCC version:
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


puneetse:
Thanks for writing this up. I'm thinking adding this content to the tutorial is going to be distracting because it is duplicating functionality/information of the nvidia-installer. From the CL side this mismatch should not be happening and the nvidia-installer does the check, blocks accordingly, and provides the same info. So I feel having users do the check manually is redundant.

lebensterben:
Indeed the check is redundant. 

But without this, users would spend quite some time applying workarounds and making modifications to their system, and only to find out the GCC version mismatch when they install the driver. Then they may spend the same amount of time restoring their system back to default.

This check avoids this kind of situation and improves user experience.

puneetse:
Let's leave this out of the documentation for the time being since it should not be happening in the first place. I do think it's helpful that you put it in your scripts. If this become a recurring problem, we can look at resurrecting this PR.

mvincerx:
@lebensterben, you could also publish these steps on [the forum](https://community.clearlinux.org/). 
BTW: These are super nice regex captures.

PR #1091, opened 2020-03-24
Remove CBM "update" after "set-timeout"

### LINKED ISSUE #1090: Update docs. containing CBM "update"
"update" is no longer needed after "set-timeout" and "set-kernel".

bktan8:
Closes #1090

### DIFF
--- a/source/guides/maintenance/enable-systemd-boot-menu.rst
+++ b/source/guides/maintenance/enable-systemd-boot-menu.rst
@@ -10,7 +10,6 @@ editing kernel command line parameters, etc.
 
 To set a timeout value for the systemd-boot menu, follow these steps:
 
-
 #. Boot up |CL|.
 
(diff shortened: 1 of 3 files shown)


mvincerx:
LGTM. 

So has clr-boot-manager been improved to handle updates automatically?

bktan8:
@mvincerx - CBM used to require running "update" for the change such as "set-timeout" or "set-kernel" to take effect.  Now, it's automatic.  But "update" is still used whenever you add a new kernel and you want CBM to know about it.

mvincerx:
> @mvincerx - CBM used to require running "update" for the change such as "set-timeout" or "set-kernel" to take effect. Now, it's automatic. But "update" is still used whenever you add a new kernel and you want CBM to know about it.

Thanks for the explanation. 

LGTM.

PR #1094, opened 2020-03-24
Minor updates to the Stateless doc.

### LINKED ISSUE #1089: reference of  the file `/usr/share/defaults/httpd/httpd.conf`  in the stateless page is not present on a live CLR OS

In the stateless document, this line points to the following file `/usr/share/defaults/httpd/httpd.conf`

https://github.com/clearlinux/clear-linux-documentation/blob/master/source/guides/clear/stateless.rst#id28

but looking into this file on today's (`32660`) CLR OS, file does not exit. My recommendation is to pick another conf file as reference.

```
# cat /usr/lib/os-release
NAME="Clear Linux OS"
VERSION=1
ID=clear-linux-os
ID_LIKE=clear-linux-os
VERSION_ID=32660
PRETTY_NAME="Clear Linux OS"
ANSI_COLOR="1;35"
HOME_URL="https://clearlinux.org"
SUPPORT_URL="https://clearlinux.org"
BUG_REPORT_URL="mailto:dev@lists.clearlinux.org"
PRIVACY_POLICY_URL="http://www.intel.com/privacy"
BUILD_ID=32660
```
**Additional context**

bktan8:
Closes #1089

### DIFF
--- a/source/guides/clear/stateless.rst
+++ b/source/guides/clear/stateless.rst
@@ -4,7 +4,7 @@ Stateless
 #########
 
 In most operating systems, user data, system data, and configuration files
-can become intermingled.
+can become intermingled, which can make them challenging to manage.
 
 .. figure:: figures/stateless-1.png
    :scale: 45%
@@ -24,7 +24,7 @@ ephemeral or non-persistent.
 File-level separation
 *********************
 
-To accomplish a stateless design the Linux Filesystem Hierarchy is separated
+To accomplish a stateless design, the |CL| filesystem hierarchy is separated
 between user-owned areas and |CL|-owned areas.
 
 .. figure:: figures/stateless-2.png
@@ -63,16 +63,17 @@ Default configurations
 ======================
 
 Software in |CL| provides default configuration values so that it is
-immediately functional, whenever it is appropriate to do so.
+immediately functional, except for some that require additional configuration.
 
-|CL| distributed software packages may be directly modified to include default
-configuration values or default configuration files may be provided by |CL|
-under :file:`/usr/share/defaults`. These files can be referenced as templates
-for customization.
+If an upstream software puts default configurations in multiple locations 
+such as :file:`/usr/` and :file:`/etc`, it will be modified by the |CL| 
+distro to comply with the stateless design.  Also, some default configurations 
+may be modified to close security loopholes.  Defaults will reside 
+under :file:`/usr/share/defaults`.  These files can be referenced as 
+templates for customization.
 
-For example, the default configuration that Apache uses when installed can be
-found at :file:`/usr/share/defaults/httpd/httpd.conf` directory.
-
+For example, after installing the `httpd` bundle for Apache web server, its 
+default configurations appear in the :file:`/usr/share/defaults/httpd/` directory.
 
 Overriding configurations
 =========================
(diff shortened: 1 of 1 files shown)


source/guides/clear/stateless.rst
```diff
@@ -65,14 +65,14 @@ Default configurations
...
 configuration values or default configuration files may be provided by |CL|
 under :file:`/usr/share/defaults`. These files can be referenced as templates
 for customization.
 
-For example, the default configuration that Apache uses when installed can be
-found at :file:`/usr/share/defaults/httpd/httpd.conf` directory.
-
+As an example, default configurations that the Apache web server uses,
+after installing the `httpd` bundle, can be found in the 
+:file:`/usr/share/defaults/httpd/` directory.
```
mvincerx: For example, if you want to use the Apache web server, install the `httpd` bundle, as shown in the next section.  

[I think it's best to keep this simple. No need to explain what you're already showing how to do in next section.]

bktan8: I see them as 2 separate topics.  The first is "defaults".  The 2nd is how to override.  
it just so happens that we're using Apache as the example in both topics.

mvincerx: OK. I can live with that. I see your point.  How about this: 

For example, after installing the `httpd` bundle for Apache web server, default configurations appear in :file:`/usr/share/defaults/httpd/` directory.


I found it hard to understand the writer's intent.

bktan8: done.

mvincerx: Thank you.

mvincerx:
@puneetse, please review.

source/guides/clear/stateless.rst
```diff
@@ -65,14 +65,13 @@ Default configurations
 Software in |CL| provides default configuration values so that it is
 immediately functional, whenever it is appropriate to do so.
 
-|CL| distributed software packages may be directly modified to include default
+|CL| distributed software may be directly modified to include default
```
puneetse: Consider rephrasing this to emphasize that |CL| modifies the software. Right now it could be interpreted as the user may modify...

mvincerx:
LGTM.
Thanks, all. :)

source/guides/clear/stateless.rst
mvincerx: Good point. I'll wait on @bktan8.

bktan8: How's this?
```
Software in |CL| provides default configuration values so that it is
immediately functional, whenever it is appropriate to do so.

If an upstream software puts default configurations in multiple locations such as :file:`/usr/` and :file:`/etc`, it will be modified by the |CL| distro to comply with the stateless design.  Also, some default configurations may be modified to close security loopholes.  The defaults will reside under :file:`/usr/share/defaults`.  These files can be referenced as templates
for customization.
```

puneetse: Nicely worded.

You could simplify by removing this sentence:
> Also, some default configurations may be modified to close security loopholes.

with this in the first sentence:

> Software in |CL| provides *secure* default configuration values so that it is

mvincerx: Agreed. Nice job, @bktan8. I prefer the first draft on security loopholes because it addresses the importance of integrating CVE patches from upstream. 

I don't understand the phrase, "whenever it is appropriate to do so."

In my reading, the sentence stands strong without that phrase. 
Was a shade of meaning lost by removing it?

bktan8: @puneetse - I'd like to keep the first sentence as is.  The sentence about security closing loopholes is needed in order to emphasize that we purposely do that.  Simply adding "secure" to the first sentence, I'm afraid, will get lost in the messaging.

@mvincerx - regarding "whenever it is appropriate to do so" - some software we distribute isn't immediately functional.   The user has to do additional configuration to get it running.  For example, nginx.  After installing the bundle, you don't want to start it automatically because you have to configure it for your specific web server.  So it's "whenever it's appropriate to do so."

mvincerx: Oh, sure. In that case, I recommend using some of your response above. 

"...immediately functional,  except some bundles require additional configuration.

I'm open to your version, too. I just found the phrase "whenever it is appropriate to do so" disconnected from the first clause of the sentence. Wild goats eat green grass. You know. That kind of non-sequitur.

bktan8: ok - all suggested incorporated.

mvincerx: LGTM. 

@puneetse?

mvincerx:
Merging. Thanks,  all.

PR #1121, opened 2020-04-15
Simple HPC tutorial

### LINKED ISSUE #1080: Tutorial: Simple Computer Cluster on Clear Linux for IA
Step-by-step instructions to install Slurm, Munge and PDSH
in order to manage jobs across a computer cluster consisting of a controller
node and additional worker nodes.

This was started by Juro Bystricky.

bktan8:
Closes #1080


source/tutorials/hpc.rst
```diff
@@ -0,0 +1,688 @@
...
+
+#. Permit the root user to SSH.  This is necessary for password-less SSH setup 
+   later.
+
+   .. code-block:: bash
+
+      echo "PermitRootLogin yes" | sudo tee -a /etc/ssh/sshd_config 
+      sudo systemctl restart sshd
+
+#. Set the hostname to `hpc-worker` plus its number.
```
mvincerx, line 163: "its number"? Where is its number found? Please be more specific and reference this number beforehand.

bktan8: As described at the intro section, this tutorial involves 4 worker nodes.  Their hostnames are "hpc-worker1", "hpc-worker2", "hpc-worker3", and "hpc-worker4".  Those are the numbers.

mvincerx:
@jurobystricky Will you please help review this PR?

source/tutorials/hpc.rst
mvincerx: Revision makes it much clearer. Thanks.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,688 @@
...
+
+      sudo swupd update
+
+#. Set the password for the root user.
+
+   .. code-block:: bash
+
+      sudo passwd root
+
+#. Set the hostname of the system to `hpc-controller`.
```
mvincerx, line 58: Changing the hostname can have potentially negative effects. For example, if changing hostname remotely, this might cause problems with `sshd.service` and `NetworkManager-wait-online.service`, which could disable the connection. 

Please add some type of warning here--doesn't have to be specific as above--but users should understand this risk before executing commands.

bktan8: Can you point me to the source of this information?  hostnamectl is mostly setting the "pretty" name and shouldn't affect routability.  Perhaps, this step can be omitted to avoid confusion.

mvincerx: The source is my tests. Will set up a call to discuss.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,688 @@
...
+
+.. figure:: /_figures/hpc/hpc-01.png
+   :alt: Simple HPC cluster
+
+   Figure 1: Simple HPC cluster
+
+The configuration is intentionally kept simple, notably avoiding setting 
+up cgroups and accounting. These and many more additional configuration 
+options can be added later.
+
```
mvincerx, line 23:
```suggestion
.. note::
   This tutorial assumes you start with a new installation of |CL|.
```

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,688 @@
...
+Prerequisites
+*************
+
+* Knowledge and experience with HPC 
+* Familiarity with slurm, munge, and PDSH
+* All nodes have synchronized clocks (typically by NTP)
+
```
mvincerx, line 30:
```suggestion
.. caution::
   Modifying the `hostname` of a pre-existing host may potentially cause networking errors.   
   Therefore, we recommend starting with a new installation to achieve the best results.
```

bktan8: Since we also set the assumption of fresh installation at the intro, I do not think this "caution" is necessary.

mvincerx: That's fine.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,689 @@
...
+.. figure:: /_figures/hpc/hpc-01.png
+   :alt: Simple HPC cluster
+
+   Figure 1: Simple HPC cluster
+
+The configuration is intentionally kept simple, notably avoiding setting 
+up cgroups and accounting. These and many more additional configuration 
+options can be added later.
+.. note::
+   This tutorial assumes you start with a new installation of |CL|.
```
mvincerx, line 24: Add an empty line after Ln 24.

mvincerx:
I think it's worth waiting for a review from @jurobystricky, too.

jurobystricky:
Yes, please wait, I have some input as well. Will try to do it today, at the latest tomorrow.

bktan8:
Take your time @jurobystricky.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+.. contents::
+    :local:
+    :depth: 1
+
+Set up controller node
+**********************
+
+In this step, install the cluster tools, enable the slurm controller service, 
+and configure and enable the the munge service.
+
```
jurobystricky, line 45: The munge service should be installed/started before any other service (slurmctld, slurmd,..)
The Slurm systemd service unit files have : "After=network.target munge.service"

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+   .. code-block:: bash
+
+      sudo passwd root
+
+#. Set the hostname of the system to `hpc-controller`.
+
+   .. code-block:: bash
+
+      sudo hostnamectl set-hostname hpc-controller
+
```
jurobystricky, line 68: Should be mentioned we don't really need to change the existing controller hostname, we do it only because the steps in this document assume this name.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+#. Install the `cluster-tools` bundle.
+
+   .. code-block:: bash
+
+      sudo swupd bundle-add cluster-tools
+
+#. Start the slurm controller service and enable it to start automatically 
+   on boot.
+
```
jurobystricky, line 83: Please reverse the order: first step should be start the munge key/service, then slurmctld.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      sudo passwd root
+
+#. Permit the root user to SSH.  This is necessary for password-less SSH setup 
+   later.
+
+   .. code-block:: bash
+
+      echo "PermitRootLogin yes" | sudo tee -a /etc/ssh/sshd_config 
+      sudo systemctl restart sshd
+
```
jurobystricky, line 167: You need to create the folder /etc/ssh first, then create the file /etc/ssh/sshd_config. On a freshly installed CLR they don't exist.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+#. Log into the controller node with the regular user account.
+
+#. Generate an SSH authentication key pair for the regular user.
+
+   .. code-block:: bash
+
+      ssh-keygen -t rsa
+
+   Accept all default options.
+
```
jurobystricky, line 205: Maybe stress the fact that in particular we don't want any passphrase (it should be the default)

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      sudo mkdir -p /etc/slurm
+      sudo cp -v /usr/share/defaults/slurm/slurm.conf /etc/slurm
+
+#. Add the controller information.
+
+   a. :command:`sudoedit` the :file:`slurm.conf` file. Set the `ControlMachine`
+      value to the controller's hostname and set the `ControlAddr` to its 
+      IP address. 
+
```
jurobystricky, line 313: Strictly speaking, controller IP address is not really necessary. But as you have just set up a new new hostname, it may take some time for DHCP/DNS to resolve the name. So in this context it is possible that without the IP addresses you could have problems. But it also assumes the IP addresses don't change.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+	 hpc-worker3
+	 hpc-worker4
+	 EOF
+
+   #. Using PSDH, get the hardware configuration of each node.
+
+      .. code-block:: bash
+
+         pdsh -w ^worker-nodes-list slurmd -C
+
```
jurobystricky, line 341: Just an observation: pdsh uses ssh by default and accesses the workers by hostnames here, no worker IP address needed. (See comments below)

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      .. code-block:: console
+
+         # 
+         # COMPUTE NODES (mode detailed values reported by "slurmd -C" on each node)
+         NodeName=hpc-worker1 NodeAddr=192.168.1.20 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=648
+         NodeName=hpc-worker2 NodeAddr=192.168.1.21 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=648
+         NodeName=hpc-worker3 NodeAddr=192.168.1.22 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=1497
+         NodeName=hpc-worker4 NodeAddr=192.168.1.23 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=1723
+
```
jurobystricky, line 393: Again, the IP addresses are optional. You already used pdsh to access the workers by their hostnames via the workers-nodes-list, so the name resolution must have worked by now.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      .. tip::
+
+         It is strongly recommended to set the `RealMemory` value for each 
+	 worker node slightly below, say 90%, what was reported by 
+	 :command:`slurmd -C` 
+	 in case some memory gets use by some processes, which would 
+	 cause slurm not to make a node not available due to its memory 
+	 resource falling below the stated value in the configuration file.    
+
```
jurobystricky, line 382: Did you mean "which would  cause Slurm to make a node not available" ?  There are two negatives in the sentence, so it is a bit confusing (to me).

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+         NodeName=hpc-worker2 NodeAddr=192.168.1.21 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=648
+         NodeName=hpc-worker3 NodeAddr=192.168.1.22 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=1497
+         NodeName=hpc-worker4 NodeAddr=192.168.1.23 CPUs=1 Boards=1 SocketsPerBoard=1 CoresPerSocket=1 ThreadsPerCore=1 RealMemory=1723
+
+   #. Create partitions.
+
+      A slurm partition is basically the grouping of worker nodes.  
+      Give each partition a name and decide which worker node(s) go belong to 
+      it.
+
```
jurobystricky, line 399: ..which worker node(s) belong to it.. -> excessive "go"

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      For example:
+   
+      .. code-block:: console
+      
+         PartitionName=workers Nodes=hpc-worker1,hpc-worker2,hpc-worker3, hpc-worker4 Default=YES MaxTime=INFINITE State=UP
+         PartitionName=debug Nodes=hpc-worker1,hpc-worker3 MaxTime=INFINITE State=UP
+
```
jurobystricky, line 406: Formatting issue: This is a bit misleading, should be split in two separate examples. User can use either of the two partitions definitions, but not both of them. Also an extra space following ',' in front of hpc-worker4

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+   #. Set the ownership of the :file:`slurm.conf` file to slurm.
+
+      .. code-block:: bash
+         
+	 sudo chown slurm: /etc/slurm/slurm.conf
+
+Copy :file:`slurm.conf` and munge key to all worker nodes
+*********************************************************
+
```
jurobystricky, line 420: I am pretty sure you need to start munge service before slurmd service on each node. So the order should be reversed. If I'm wrong then I just learned something new. Please see : https://slurm.schedmd.com/quickstart_admin.html 
Quoting:
 Make sure that all nodes in your cluster have the same munge.key. Make sure the MUNGE daemon, munged is started before you start the Slurm daemons.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+         sudo scontrol update nodename=<worker> state=idle reason=""
+
+#. And finally, verify slurm can run jobs on all 4 worker nodes by issuing
+   a simple :command:`hostname` command.
+
+   .. code-block:: bash
+
+      srun -N4 -p worker hostname
+
```
jurobystricky, line 546: -p workers, missing 's' from the partition name.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+   #SBATCH --output=show-hostnames.out
+   #
+   #SBATCH --ntasks=4
+   #SBATCH --time=10:00
+   #SBATCH --mem-per-cpu=100
+   #SBATCH --ntasks-per-node=1
+
+   srun hostname
+   EOF
+
```
jurobystricky, line 577: This does not display correctly in my browser, should it be in a code block?
.. code-block:: bash

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+#. Execute the script.
+
+   .. code-block:: bash
+
+      sbatch show-hostnames.sh
+   
+   The result will be appear on first node.
```
jurobystricky, line 584: Redundant  "be", more precisely:
The result will appear on the first node of the used partition. As no partition was explicitly specified, it would be the default partition.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+      #!/bin/sh
+      #SBATCH -o mpi-helloworld.out
+      #SBATCH --nodes=4
+      #SBATCH --ntasks-per-node=1
+
+      srun ./mpi-helloworld
+      EOF
+
+#. View the results on first worker node.
+
```
jurobystricky, line 672: As above:  the first worker node in the partition.

source/tutorials/hpc.rst
mvincerx: 👍

source/tutorials/hpc.rst
mvincerx: @jurobystricky also commented on this below.

source/tutorials/hpc.rst
```diff
@@ -0,0 +1,693 @@
...
+A Simple HPC Cluster on |CL|
+############################
+
+This tutorial demonstrates how to set a simple :abbr:`HPC (High
+Performance Computing)` cluster using `slurm`_, `munge`_, and 
+`PDSH`_.  For this tutorial, this cluster consists of a controller node 
+and four worker nodes, as shown in Figure 1.  For the sake of simplicity,
+each node resides on a separate host and their hostnames are hpc-controller,
+hpc-worker1, hpc-worker2, hpc-worker3, and hpc-worker4.  
+
```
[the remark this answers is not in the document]
bktan8: done.

source/tutorials/hpc.rst
bktan8: Actually /etc/ssh exists after a fresh installation.  It's an empty folder, but it's there.  I confirmed on both the desktop and server installation.

source/tutorials/hpc.rst
bktan8: done.

source/tutorials/hpc.rst
bktan8: Ok - I will remove the IP address portion for both the controller and nodes from the document.

source/tutorials/hpc.rst
bktan8: yep.  I removed the IP address portion in the slurm.conf file.

source/tutorials/hpc.rst
bktan8: Good catch on the first "not".

source/tutorials/hpc.rst
bktan8: Good catch!

source/tutorials/hpc.rst
bktan8: writer error!  fixed.

source/tutorials/hpc.rst
bktan8: Great catch!

source/tutorials/hpc.rst
bktan8: Writer error again!  Thanks!

source/tutorials/hpc.rst
bktan8: I used your version.  Thanks!

source/tutorials/hpc.rst
bktan8: @jurobystricky - Out of curiosity, does the order the workers in a partition matter?

For example: 
PartitionName=workers Nodes=hpc-worker1, hpc-worker2, hpc-worker3 Default=YES MaxTime=INFINITE State=UP

vs. 

PartitionName=workers Nodes=hpc-worker2, hpc-worker1, hpc-worker3 Default=YES MaxTime=INFINITE State=UP

In the first instance, the first worker is hpc-worker1.
In the 2nd instance, the first worker is hpc-worker3.

source/tutorials/hpc.rst
bktan8: Ok.  Order changed.

source/tutorials/hpc.rst
bktan8: done.

source/tutorials/hpc.rst
bktan8: It seemed to work the way I had it.  But I'll take the advice of starting MUNGE before Slurm daemons.

bktan8:
@jurobystricky - would you mind taking a look at the latest?

bktan8:
@jurobystricky - Do you have any feedback before we publish?

mvincerx:
@jurobystricky, thanks for reviewing. Do you have any more feedback?

jurobystricky:
Looks good. We can add a more complex example(s) later on.

mvincerx:
LGTM. Thanks, all for the intensive feedback and testing.

bktan8:
Thank you MV!

PR #1152, opened 2020-05-04
Update list of display managers on about page

puneetse:
This PR updates the list of display managers available in the about page to align with the announced refocusing. 

Closes https://github.com/clearlinux/clear-linux-documentation/issues/1158

### DIFF
--- a/source/about.rst
+++ b/source/about.rst
@@ -47,8 +47,13 @@ build a unique Linux distro. Developing a distro in house allows us to experimen
 
 We leverage the pool of knowledge and skills at Intel to drive improvements to |CL|.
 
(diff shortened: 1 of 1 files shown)


source/about.rst
```diff
@@ -296,4 +296,4 @@ Which Components are used in Clear Linux?
 .. _Supplied Bundles: https://clearlinux.org/software
 .. _micro-config-drive: https://github.com/clearlinux/micro-config-drive
 .. _Telemetrics: https://github.com/clearlinux/telemetrics-backend
-.. _packages source code: https://github.com/clearlinux-pkgs/
\ No newline at end of file
+.. _packages source code: https://github.com/clearlinux-pkgs/
```
mvincerx, line 299: Add 

```
.. _Clear_linux_store: https://clearlinux.org/software

```

puneetse: It is already there, there was no space between "see" and the link in the original copy causing it to not get picked up. Fixed by adding the space.

mvincerx:
LGTM.

PR #1157, opened 2020-05-05
Adds support for sphinx_copybutton.

### LINKED ISSUE #1134: Create a Sphinx copybutton for code-blocks.
**Do you think Clear Linux documentation needs a new document? Please describe.**

Create a Sphinx copybutton for code-blocks. 
https://pypi.org/project/sphinx-copybutton/
https://sphinx-copybutton.readthedocs.io/en/latest/

**Should the new document be a guide, a reference, or a tutorial?**
NA. Applies across all documents. 

**Describe or provide examples of similar documents, if possible, from other web sites**
NA

**Additional context**
NA

mvincerx:
Closes #1134

IP

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ sphinx-sitemap==1.0.2
 Jinja2==2.10.1
 GitPython==3.0.8
 sphinx-tabs
+sphinx-copybutton
--- a/source/conf.py
+++ b/source/conf.py
@@ -38,6 +38,10 @@ extensions = [
     'sphinx_tabs.tabs', 'sphinx_copybutton'
 ]
 
+#copybutton_prompt_text = "#"
+
+#copybutton_remove_prompts = False
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
 
--- a/source/conf.py
+++ b/source/conf.py
@@ -38,10 +38,6 @@ extensions = [
     'sphinx_tabs.tabs', 'sphinx_copybutton'
 ]
 
-#copybutton_prompt_text = "#"
-
-#copybutton_remove_prompts = False
-
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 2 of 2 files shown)


intelkevinputnam:
Looks like the requirements.txt needs to be updated.

mvincerx:
Yep. Testing other stuff too.

intelkevinputnam:
Based on a conversation I'm having with the PMEM-CSI folks, maybe we should seriously consider using "console" with both input/output and marking input lines with "$". We could use copybutton_prompt_text = "$ " to avoid copying output and the prompt when the clicking the button. What do you think?

mvincerx:
How about sharing your thoughts on why PMEM-CSI prefers this in our meeting tomorrow?

I'm open to this, but there are a few things to work out.  For example, a knock-on effect is that the hash represents root, like: `root@yourbox:~#`. For some commands that require showing root (docker, stacks, etc.), `sphinx-copybutton` would capture the `#` too. But we don't want that. If you have a solution to address that, please share. 

If we adopted using `console` for both input/output, it would mean reverting to the way CL Docs used to represent code-blocks about 2 years ago.  There was a lot of debate about it back then, so we should revisit this now as a team.

mvincerx:
Like we discussed in the Docs meeting, go ahead an push directly to this branch. I'll hold off.

intelkevinputnam:
@mvincerx Can you verify that you're allowing pushes?

mvincerx:
Verified.

intelkevinputnam:
K. I'm getting permission denied when I try to push to the mv-copybutton branch of your fork.

mvincerx:
@intelkevinputnam Looking into this. I'm not sure why you receive permission denied. It's certainly not intentional.

intelkevinputnam:
@mvincerx No worries. It is pretty weird as I still have merge permissions.

PR #1162, opened 2020-05-07
Escapes local host link, so it isn't turned into active link.

intelkevinputnam:


mvincerx:
LGTM.

PR #1174, opened 2020-05-18
peteonrails/add zfs tutorial

peteonrails:
This PR introduces a new tutorial for ZFS. I have read the contribution guidelines and have tried to adhere to them -- feedback welcome.

I stopped before adding a ZFS-on-root section because this tutorial is already long. ZFS-on-root is substantially harder, so I would welcome feedback about whether it should be a section of this tutorial, or a tutorial all of its own.

I hope this PR will help the many users who ask the core team weekly about adding a ZFS driver to Clear Linux. For that reason, the background section is a little bit longer than most of the other tutorials. I could see a case for trimming that content, but I think it helpful to reduce questions int he future.

I've contributed to open source projects before, so I expect to run a gauntlet of feedback here.

I am subscribed to: dev@lists.clearlinux.org
My Forum username: peteonrails 

Thanks in advance for considering this request.

### DIFF
--- a/source/tutorials/zfs.rst
+++ b/source/tutorials/zfs.rst
@@ -3,17 +3,12 @@
 ZFS
 ###
 
-This tutorial is not quite ready for prime-time. I think
-a reasonably skilled Linux user will be able to use these
-steps to get up and running with ZFS on Clear.
-
-If you use this document, YMMV and you might break stuff. Be
-careful.
+This tutorial covers the setup of ZFS-on-Linux under Clear Linux, 
+using a non-root device for your zpools. ZFS on root is a work-in-progress. 
 
 .. contents:: 
    :local:
    :depth: 2
-
 
 Background
 **********
@@ -30,9 +25,8 @@ Some distributions of Linux take a different read of the
 CDDL / GPL issue -- most notably Canonical's Ubuntu -- and
 argue that distributing a binary kernel module alongside Linux
 does not infringe on the GPL nor the CDDL, since no derivative
-work has been created. The Clear Linux team does not share
-that optimistic view, and therefore zfs.ko must be built by
-the user under Clear Linux.
+work has been created. Clear Linux does not share that view, 
+and therefore zfs.ko must be built by the user under Clear Linux.
 
 To read the argument against shipping zfs.ko along with Linux:
 https://sfconservancy.org/blog/2016/feb/25/zfs-and-linux/
@@ -71,41 +65,12 @@ a lag between a new kernel release and a supprted ZFS driver.
 Installation Types
 ******************
 
-Root installations
-==================
+ZFS can be easily installed for use on non-root devices. It is substantially
+harder to get ZFS to work on a root partition, but it is possible if that's what
+you want. 
 
-Since there is a lag between kernel releases and ZFS support, I recommend
-that you use a long term support kernel along with the latest ZFS driver
-if you are going to run ZFS on root.
-
-If you choose to use a native kernel and ZFS on root, clr-boot-manager
-will refuse to update your kernels, so you will have to get comfortable
-with installing new kernels to systemd-boot, which is a good skill to
-have in any case.
-
-We will explore ZFS on root after we get a working ZFS driver for non-root disks.
-
-Non-Root installation
-=====================
-
-Before making the decision to run your CLR_ROOT partition
-on a ZFS dataset, you should verify that you can get ZFS working
-for a data-only partition.
-
-If you use ZFS on a partition other than CLR_ROOT, with a root
-partition that is ext4, xfs, or f2fs you may run into problems mounting
-ZFS when a new kernel is released (until I have DKMS working).
-
-In this configuration, clr-boot-manager will install your new kernel,
-which may or may not work with zfs.ko.
-
-- DKMS may not have rebuilt the module
-- DKMS may not have autoinstalled the module
-- The new kernel might introduce breaking changes that prevent zfs from compiling
-
-You may end up having to manually recompile zfs.ko with the new kernel code, and zfs *might*
-not compile at all with the new kernel. **So, be sure you don't put anything on that ZFS pool that you would need
-in order to rebuild kernel modules.**
+A non-root build and installation is a prerequisite to a root installation, so 
+we will cover the non-root install first. 
 
 Prerequisites
 *************
@@ -356,30 +321,39 @@ Then create the configuration file with:
 
 When you reboot, zfs should be loaded by the kernel automatically.
 
-Using on a Non-Root device
-**************************
 
 You're ready to create zpools and datasets.
 
-Enable automatic import of a pool by zfs-import-cache.service with:
+Potential Issues on a Non-Root device
+*************************************
+When the Clear Linux kernel is upgraded, DKMS will attempt to rebuild your
+zfs module for use with the new kernel. If you boot a new kernel and cannot
+find your zpools:
 
-.. code-block:: bash
+- DKMS may not have rebuilt the module
+- DKMS may not have autoinstalled the module
+- The new kernel might introduce breaking changes that prevent zfs from compiling
 
-   zpool set cachefile=/etc/zfs/zpool.cache <pool>
+You may end up having to manually recompile zfs.ko with the new kernel code, and zfs *might*
+not compile at all with the new kernel. 
+
+**So, be sure you don't put anything on that ZFS pool that you would need
+in order to rebuild kernel modules.**
 
 ZFS on root (/)
 ***************
-WIP
 
-Installing new kernels with ZFS root   systemctl enable zfs-import-cache
-   systemctl enable zfs-import.target
+Since there is a lag between kernel releases and ZFS support, I recommend
+that you use a long term support kernel along with the latest ZFS driver
+if you are going to run ZFS on root.
+
+If you choose to use a native kernel and ZFS on root, clr-boot-manager
+will refuse to update your kernels, so you will have to get comfortable
+with installing new kernels to systemd-boot, which is a good skill to
+have in any case.
+
+Installing new kernels with ZFS root   
 ====================================
-If you use ZFS for your CLR_ROOT, then clr-boot-manager will no longer
-automatically install new kernels for you: you'll have to set them up
-manually when updated kernels are available.
-
-Hopefully, by now you understand why *this is a good thing*.
-
 When a new kernel is available, you will find that the Clear Linux tools will refuse to install your new kernel
 with an error similar to this:
 
--- a/source/tutorials/zfs.rst
+++ b/source/tutorials/zfs.rst
@@ -12,7 +12,7 @@ This tutorial shows how to set up `OpenZFS* file system and volume manager`_ on 
 Background
 **********
 
-|CL| not ship with a binary ZFS kernel module (zfs.ko). |CL| users who wish to incorporate the zfs.ko kernel module must build and maintain this work themselves.
+The OpenZFS storage platform provides volume management, snapshot capabilities, and redundancy detection. |CL| not ship with a binary ZFS kernel module (zfs.ko). |CL| users who wish to incorporate the zfs.ko kernel module must build and maintain this work themselves.
 
 .. CAUTION::
 
@@ -21,7 +21,7 @@ Background
 Known Issues
 ************
 
-Using a long-term-support (LTS) kernel when running OpenZFS reduces the risk of incompatibilities with an LTS kernel. When new kernels or new versions of OpenZFS are released, users bear the responsibility to test those releases and ensure compatibility before deploying any updates. 
+Using a long-term-support (LTS) kernel when running OpenZFS reduces the risk of incompatibilities with kernel updates. When new kernels or new versions of OpenZFS are released, users bear the responsibility to test those releases and ensure compatibility before deploying any updates. 
 
 Prerequisites
 *************
@@ -59,7 +59,7 @@ If you do not currently use a DKMS kernel, install it by using one of the option
 Bundles
 =======
 
-Before you install OpenZFS, install these bundles.
+Before you install OpenZFS, install the bundles that contain the build dependencies.
 
 .. code-block:: bash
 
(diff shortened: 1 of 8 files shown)


mvincerx:
Thanks, @peteonrails . You can expect a status update in the next day.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+GPL are not incompatible with respect to redistribution of
+source code.
+
+There is nothing stopping a user from downloading the ZFS drivers,
+compiling them, then installing them for their own use. This is no
+different from downloading and using proprietary drivers like the Nvidia
+Graphics or Broadcom wireless drivers.
+
+Since the binaries for ZFS cannot be distributed with the
+Linux kernel binaries, you'll have some work to  ensure compatbility
```
bwarden, line 46:
```suggestion
Linux kernel binaries, you'll have some work to ensure compatibility
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+when new kernels are released.
+
+This is not always trivial, since new kernel release candidates will
+invariably break *something* in ZFS. The ZFS-on-Linux team is
+very good about catching up quickly, but on occasion you will find
+a lag between a new kernel release and a supported ZFS driver.
```
bwarden, line 52: We may want to recommend using an LTS kernel here, to reduce the likelihood of breaking changes between the kernel and ZFS.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+
+DKMS kernel
+===========
+
+This tutorial assumes you are using a DKMS kernel. You should read and
+understand how to :ref:`kernel-modules-dkms`.
+
+If you do not currently use a DKMS kernel, install it with the following steps.
+
+Check whether you have an lts or native kernel:
```
bwarden, line 67:
```suggestion
Check whether you have an LTS or native kernel:
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+
+    uname -r
+
+If 'native' appears in the kernel name, then install a native kernel with DKMS support:
+
+.. code-block:: bash
+
+    sudo swupd bundle-add kernel-native-dkms
+
+If 'lts' appears in the kernel name, then install the latest TLS kernel with DKMS support:
```
bwarden, line 79:
```suggestion
If 'lts' appears in the kernel name, then install the latest LTS kernel with DKMS support:
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+.. code-block:: bash
+
+    sudo swupd bundle-add linux-dev
+
+If you are using an LTS kernel:
+
+.. code-block:: bash
+
+    sudo swupd bundle-add linux-lts-dev
+
```
bwarden, line 103: The `-dev` bundles are automatically included by the corresponding `-dkms` bundles, so they don't need to be added explicitly here.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+Finally, install need these bundles, no matter which kernel you are using:
```
bwarden, line 104:
```suggestion
Finally, install these bundles, no matter which kernel you are using:
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+    sudo swupd bundle-add os-core-dev devpkg-openssl devpkg-util-linux
+
+Install
+*******
+
+Get the ZFS code
+================
+
+The ZFS codebase moves quickly, just like the kernel codebase. Therefore
+it's best that you get the source code from the ZFS on Linux repository,
```
bwarden, line 117: Maybe make "ZFS on Linux" a hyperlink?

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+hosted on Github.
+
+Fetch the ZFS repository via git:
+
+.. code-block:: bash
+
+   git clone https://github.com/openzfs/zfs.git /opt/src/zfs
```
bwarden, line 124: You'll need a `mkdir` here.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Remember where you check-out the git repository, because you will need it
```
bwarden, line 126:
```suggestion
Remember where you clone the git repository, because you will need it
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+zfs-volume-wait.service
+```
+
+If you want to use ZFS automatic zpool import and filesystem
+mount services, copy the systemd.unit files into /etc and enable them:
+
+.. code-block:: bash
+
+   sudo cp ./etc/systemd/system/*.service /etc/systemd/system/
+   sudo cp ./contrib/dracut/90zfs/zfs-env-bootfs.service /etc/systemd/system/
```
bwarden, line 236: Might be better to symlink these?

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+   systemctl enable zfs.target
+
+If you prefer to use the zfs-mount-generator instead of zfs-mount, refer to the
+Arch Linux Guide's `ZFS Mount Generator`_ section for details. 
+
+Load the kernel module at boot
+==============================
+
+The ZFS module will not load automatically at boot. Load the zfs.ko module
+at boot time by with systemd.
```
bwarden, line 251:
```suggestion
at boot time with systemd.
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+
+First, allow unsigned modules:
+
+    echo "module.sig_unenforce" | sudo tee /etc/kernel/cmdline.d/allow-unsigned-modules.conf
+
```
bwarden, line 256: This is automatically provided by the `-dkms` bundles.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+Systemd uses the `/etc/modules-load.d/` directory to load out-of-tree kernel modules.
+Make sure that the directory exists:
+
+    sudo mkdir -p /etc/modules-load.d
+
+Create the configuration file:
+
+    sudo echo "zfs" | sudo tee /etc/modules-load.d/01-zfs.conf
```
bwarden, line 264:
```suggestion
    echo "zfs" | sudo tee /etc/modules-load.d/01-zfs.conf
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,299 @@
...
+- DKMS may not have rebuilt the module
+- DKMS may not have autoinstalled the module
+- The new kernel might introduce breaking changes that prevent zfs from compiling
+
+To fix this situation, you may have to recompile zfs.ko with the new kernel code.
+ZFS *might* not compile at all with the new kernel.
+
+**So, be sure you don't put anything on that ZFS pool that you would need
+in order to rebuild kernel modules.**
+
```
bwarden, line 283: Should add that you can check the results of dkms like this:
```
systemctl status dkms-new-kernel.service
```
or by checking the journal.

source/tutorials/zfs.rst
mvincerx: 👍

mvincerx: We may want to add that "users should test releases before deploying."

source/tutorials/zfs.rst
peteonrails: I updated to symlink just the two services we plan to use into /etc/systemd/system, and added a sentence explaining that the user should do the same for any other service units they plan to use.

source/tutorials/zfs.rst
peteonrails: Removed.

source/tutorials/zfs.rst
peteonrails: Added

source/tutorials/zfs.rst
peteonrails: Oops! Thank you. Added.

source/tutorials/zfs.rst
peteonrails: I added some language that conveys these warnings.

mvincerx:
This PR is still under internal review.

mvincerx:
Hold on pushing to this PR. I will push to this PR with approved changes.

mvincerx:
@peteonrails. Thanks again for this contribution. My latest push is what we can publish. 
@puneetse. Will you please review again for accuracy. I have not changed any commands, just wording, formatting and syntax.

source/tutorials/zfs.rst
```diff
@@ -232,51 +204,51 @@ mount services, link the systemd.unit files into /etc and enable them.
...
+.. CAUTION::
+
+   When the |CL| kernel is upgraded, DKMS will attempt to rebuild your OpenZFS module for the new kernel. 
+
+   - DKMS may not have rebuilt the module
+   - DKMS may not have autoinstalled the module
+   - The new kernel might introduce breaking changes that prevent zfs from 
+     compiling
+
+To fix this situation, you may have to recompile zfs.ko with the new kernel code.OpenZFS *might* not compile at all with the new kernel.
```
bwarden, line 240:
```suggestion
To fix this situation, you may have to recompile zfs.ko with the new kernel code. OpenZFS *might* not compile at all with the new kernel.
```

source/tutorials/zfs.rst
```diff
@@ -232,51 +204,51 @@ mount services, link the systemd.unit files into /etc and enable them.
...
 
-- DKMS may not have rebuilt the module
-- DKMS may not have autoinstalled the module
-- The new kernel might introduce breaking changes that prevent zfs from compiling
 
-To fix this situation, you may have to recompile zfs.ko with the new kernel code.
-ZFS *might* not compile at all with the new kernel.
+.. CAUTION::
+   
+   **Be sure you don't put anything on that OpenZFS pool that you would need
```
bwarden, line 245:
```suggestion
   **Be sure you don't put anything on an OpenZFS pool that you would need
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+      sudo cp -Rv /opt/src/zfs /usr/src/zfs-0.8.4
+      cd /usr/src/zfs-0.8.4
+
+#. The ZFS distribution provides a script to build a suitable dkms.conf file.
+   Build dkms.conf and install it into the DKMS tree.
+
+   .. code-block:: bash
+
+      cd /usr/src/zfs-0.8.4
+      scripts/dkms.mkconf -n zfs -v 0.8.4 -f dkms.conf
```
puneetse, line 125:
```suggestion
      sudo scripts/dkms.mkconf -n zfs -v 0.8.4 -f dkms.conf
```

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+
+      sudo cp -Rv /opt/src/zfs /usr/src/zfs-0.8.4
+      cd /usr/src/zfs-0.8.4
+
+#. The ZFS distribution provides a script to build a suitable dkms.conf file.
+   Build dkms.conf and install it into the DKMS tree.
+
+   .. code-block:: bash
+
+      cd /usr/src/zfs-0.8.4
```
puneetse, line 124: This step is duplicte with the `cd` above. One could be eliminated.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+
+If you want to use ZFS automatic zpool import and filesystem
+mount services, link the systemd.unit files into /etc and enable them.
+
+.. code-block:: bash
+
+   sudo ln -s ./etc/systemd/system/zfs-import-cache.service /etc/systemd/system/
+   sudo ln -s ./etc/systemd/system/zfs-mount.service /etc/systemd/system/
+
+   systemctl enable zfs-import-cache
```
puneetse, line 201:
```suggestion
   sudo systemctl enable zfs-import-cache
```

same suggestion for all the `systemctl` commands below.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+   systemctl enable zfs-import.target
+
+   systemctl enable zfs-mount
+   systemctl enable zfs.target
+
+OpenZFS requires you to explicitly install and enable the services you want.
```
puneetse, line 207: I suggest presenting this information earlier, towards the top of this sections header.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+   systemctl enable zfs-import.target
+
+   systemctl enable zfs-mount
+   systemctl enable zfs.target
+
+OpenZFS requires you to explicitly install and enable the services you want. 
+If you want to use other ZFS service units, symlink them similarly to the 
+example shown above.
+
+Load the kernel module at boot
```
puneetse, line 211: Is this only needed if a device in the boot path is using ZFS or for *any* ZFS filesystems to work *at all*? In other words, does it work automagically if the `mount` command is used at run time? If so, I suggest clarifying when a user would want to do this.

source/tutorials/zfs.rst
```diff
@@ -0,0 +1,275 @@
...
+
+   .. code-block:: bash
+
+      mkdir -p /opt/src/zfs
+
+#. Clone the repository.
+
+   .. code-block:: bash
+
+      git clone https://github.com/openzfs/zfs.git /opt/src/zfs
```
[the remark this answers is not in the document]
bwarden: Agreed

source/tutorials/zfs.rst
peteonrails: @mvincerx @puneetse The kernel modules must be loaded before any ZFS filesystems are mounted. If the modules are not loaded, then the `mount` command will fail with an error (I think it says: "ZFS modules not loaded.")

So, how about this: "OpenZFS kernel modules must be loaded before any OpenZFS filesystems are mounted. For convenience, load the kernel modules at boot."

mvincerx: @puneetse. I want to merge this PR. Do you have any other feedback on my latest push?

mvincerx:
Thanks, all. LGTM. This is the internally approved version.

PR #1182, opened 2020-05-29
Update TM&B in About section of docs

krisdale:
- Add disclaimers for Intel trademarks
- Add asterisk for 3rd party trademarks
- Remove redundant asterisks (only first use needed)

### DIFF
--- a/source/about.rst
+++ b/source/about.rst
@@ -241,7 +241,7 @@ Which Components are used in Clear Linux?
      - `Wayland`_ compositor
 
    * - Network services
-     - `NetworkManager`_ by default*, `systemd-networkd`_
+     - `NetworkManager`_ by default, `systemd-networkd`_ See Note below.
      - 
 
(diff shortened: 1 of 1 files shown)


krisdale:
In the last table - the networks entry - "NetworkManager by default*" What is the asterisk on 'default' trying to convey?

mvincerx:
> In the last table - the networks entry - "NetworkManager by default*" What is the asterisk on 'default' trying to convey?

Please remove the asterisk after `default`.

mvincerx:
Please hold on this, @krisdale. I'll push to make a tiny correction re your question above.

mvincerx:
> In the last table - the networks entry - "NetworkManager by default*" What is the asterisk on 'default' trying to convey?

Good catch.  I removed * and added a "See note below..." to make the connection to legacy networking on CL.

mvincerx:
LGTM.

PR #1198, opened 2020-06-30
Add one line of bash to apply flannel plugin.

inmanturbo:
A container network interface is needed to run the "manually" set up cluster without removing references to $KUBELET_NETWORK_ARGS and just using the whatever the underlying container runtime gives you. 
In most cases this is not the desired outcome.
 - Flannel is the simplest implementation
 - It works well with the instructions already given
 - This way a user can leave the "Set up Kubernetes manually" tutorial with a running cluster under their belt

### DIFF
--- a/source/tutorials/kubernetes.rst
+++ b/source/tutorials/kubernetes.rst
@@ -452,8 +452,15 @@ Kubernetes container runtime:
    .. code:: bash
 
       sudo kubeadm init \
-      --pod-network-cidr 10.244.0.0/16 
-
+      --pod-network-cidr 10.244.0.0/16
+   .. important::
+   
+      We need to install the pod network before the cluster can come up. The following will install the latest yaml file that flannel provides. 
+   
+   .. code:: bash
+   
+      kubectl apply -f https://github.com/coreos/flannel/raw/master/Documentation/kube-flannel.yml
+   
 
 #. Once the cluster initialization is complete, continue reading about how to
    `Use your cluster`_.


mvincerx:
Hey @puneetse and @eadamsintel please discuss. Thanks

inmanturbo:
Converted this to a draft because of failed checks (new to sphinx) and a couple of other reasons below:
  
- Users will also have to run
```bash
mkdir -p $HOME/.kube
sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
sudo chown $(id -u):$(id -g) $HOME/.kube/config
```
As instructed in the out put when running `kubadm init`
- May just put up a tutorial in the forum for getting up and running with `kubernetes` and `flannel`

@puneetse and @eadamsintel fel free to close this if you feel a forum topic is more appropriate.

Important things to consider: 
- The docs also give instructions for `kubadm init` using different subnets for other plugins, but do not show *how* to apply these plugins
    - should the docs then show how to apply each one in turn?

Please note the main issue being addresses here is it seems the docs should instruct how to apply at least one plugin simply and with a reasonable default (or at least how to get a single node working without one, i.e with typical Docker "bridge" networking). A good model would be one that could be linked to as a starting point to work from in the other tutorials that require a functioning kubernetes cluster, i.e [OpenFaas](https://docs.01.org/clearlinux/latest/tutorials/openfaas.html#openfaas) .

The `create_stack.sh` script  from [automatic setup](https://docs.01.org/clearlinux/latest/tutorials/kubernetes.html#set-up-kubernetes-automatically) is far too grand for this, and doesn't leave the user with any understanding as to how the stack was brought up.

Most users will want a reasonably simple and explanatory, common and functional baseline to work and build from -- i.e. a manual setup with docker and flannel. If one can get that working, one can begin writing tests to use in future projects.

eadamsintel:
@inmanturbo  When we recently updated this document we kept running into networking issues and lots of "could not connect" issues because of applying the wrong version of flannel or canal or other CNI networking plugin. We realized that with the constantly updating Kubernetes version in Clear Linux and the fact the CNI network plugins are installed from outside Clear Linux it became a real maintenance nightmare to try to match the right CNI network plugin versions (URL to a yaml file) for the different CNI plugins and match that to the Clear Linux version ranges with a particular version of Kubernetes.  That is why under related topics there is a "Installing a pod network add-on" section at the very bottom. We also put some tips on how to get past some error messages that we kept running into like "coredns pods are stuck in container creating state and logs show entries similar to one of the following" to give advice on what to do.  

I think that putting it in the forum can help keep a long running discussion that will get updated faster than this document. If it gets to a point where applying the latest CNI networking plugin from a static URL link works with multiple versions of Kubernetes going back many versions in Clear Linux so that we can document that then it would be worthwhile to do that.

We spend more effort keeping the cloud native scripts working with the latest version of Clear Linux which is why we recommended people use that as a starting point.  Those scripts actually install `Canal` as the default CNI plugin.  I think anyone who is starting a Kubernetes control plane on the command line would be advanced enough to know how to get their preferred CNI network up. If they are not that advanced then that is why we point them to the `cloud-native-basic` setup scripts.

We are thinking to modify part of the tutorial so we don't appear be recommending flannel. We are discussing internally so that we don't unintentionally mislead anyone.

inmanturbo:
@eadamsintel  that makes sense, thank you for the response.
>  If it gets to a point where applying the latest CNI networking plugin from a static URL link works with multiple versions of Kubernetes going back many versions in Clear Linux 

If that level of stability is ever reached, please mention #1198 (I'll be closing this) or put something in the newsletter to let us know! Thanks again.

eadamsintel:
Sure thing. I'll do my best to remember to contact you back.  Kubernetes keeps evolving and getting easier to initialize.

PR #1215, opened 2021-01-01
add dep for create_stack.sh kubernetes script

TopherIsSwell:
The create_stack.sh script requires git to complete fully. Perhaps this should be added to the setup_system.sh script, but until it does, this change allows the steps to complete without error.

### DIFF
--- a/source/tutorials/kubernetes.rst
+++ b/source/tutorials/kubernetes.rst
@@ -115,6 +115,14 @@ running quickly.
 
       sudo systemctl stop docker
       sudo systemctl stop containerd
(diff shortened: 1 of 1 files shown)


mvincerx:
LGTM. @TopherIsSwell Please force push to the same branch, and the CI should pass.

eadamsintel:
This LGTM also. Since the instructions state to install the `cloud-native-basic bundle` vs doing a`git clone` of the repository there is a chance git might not be installed and the `setup_system.sh` script doesn't install git.

PR #1219, opened 2021-01-31
Update lamp-server.rst

alaasalama:
sudo swupd bundle-add database-basic

needs to be replaced by:

sudo swupd bundle-add mariadb

As database-basic bundle is invalid.


mvincerx:
Fixed in https://github.com/clearlinux/clear-linux-documentation/pull/1232

PR #1250, opened 2022-09-13
Remove escaping from installer config password hashes

bwarden:
Password hashes in the installer configuration yaml are written as-is to `/etc/shadow`. The backslashes preceding `$` are unnecessary and produce an invalid hash string that will prevent login.
Removing the backslashes makes the sample password work as described.


bwarden:
Fixes #1251

pixelgeek:
Michael, can you merge this patch?

PR #1264, opened 2023-08-23
Change site name

djklimes:
Update site name to remove references to the old site.  lastestHTML to be updated from this once merged.

### DIFF
--- a/source/_themes/otc_tcs_sphinx_theme/cversions.html
+++ b/source/_themes/otc_tcs_sphinx_theme/cversions.html
@@ -21,7 +21,7 @@
       <dl>
         <dt>{{ _('clearlinux links') }}</dt>
           <dd>
-            <a href="https://https://clearlinux.github.io/">Project Home</a>
+            <a href="https://clearlinux.github.io/">Project Home</a>
           </dd>
           <dd>
             <a href="https://github.com/clearlinux/clear-linux-documentation">GitHub</a>
--- a/source/_themes/otc_tcs_sphinx_theme/cversions.html
+++ b/source/_themes/otc_tcs_sphinx_theme/cversions.html
@@ -19,9 +19,9 @@
         {% endfor %}
       </dl>
       <dl>
-        <dt>{{ _('clearlinux.org links') }}</dt>
+        <dt>{{ _('clearlinux links') }}</dt>
           <dd>
-            <a href="https://www.clearlinux.org/">Project Home</a>
+            <a href="https://https://clearlinux.github.io/">Project Home</a>
           </dd>
           <dd>
             <a href="https://github.com/clearlinux/clear-linux-documentation">GitHub</a>
--- a/source/guides/clear/security.rst
+++ b/source/guides/clear/security.rst
@@ -140,8 +140,8 @@ some examples:
 
 *Intel and the Intel logo are trademarks of Intel Corporation or its subsidiaries.*
 
-.. _`Security for software update in Clear Linux* OS`: https://clearlinux.org/blogs/security-software-update-clear-linux-os-intel-architecture
-.. _`Recent GNU* C library improvements`: https://clearlinux.org/blogs/recent-gnu-c-library-improvements
+.. _`Security for software update in Clear Linux* OS`: https://clearlinux.github.io/news-blogs/security-software-update-clear-linux-os.html
+.. _`Recent GNU* C library improvements`: https://clearlinux.github.io/news-blogs/recent-gnu-c-library-improvements.html
 .. _`rolling release`: https://en.wikipedia.org/wiki/Rolling_release
 .. _`clrtrust`: https://github.com/clearlinux/clrtrust
 .. _`Address space layout randomization (ASLR)`: https://en.wikipedia.org/wiki/Address_space_layout_randomization
--- a/source/guides/clear/swupd.rst
+++ b/source/guides/clear/swupd.rst
@@ -344,7 +344,7 @@ Related topics
 
 .. _swupd source documentation: https://github.com/clearlinux/swupd-client/blob/master/docs/swupd.1.rst
 
-.. _Kata Containers: https://clearlinux.org/downloads/containers
+.. _Kata Containers: https://clearlinux.github.io/downloads/containers
 
 .. _containers-virt: https://github.com/clearlinux/clr-bundles/blob/master/bundles/containers-virt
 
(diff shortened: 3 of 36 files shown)


source/_themes/otc_tcs_sphinx_theme/cversions.html
```diff
@@ -19,9 +19,9 @@
         {% endfor %}
       </dl>
       <dl>
-        <dt>{{ _('clearlinux.org links') }}</dt>
+        <dt>{{ _('clearlinux links') }}</dt>
           <dd>
-            <a href="https://www.clearlinux.org/">Project Home</a>
+            <a href="https://https://clearlinux.github.io/">Project Home</a>
```
bryteise: Double https://

source/guides/clear/security.rst
```diff
@@ -140,8 +140,8 @@ some examples:
 
 *Intel and the Intel logo are trademarks of Intel Corporation or its subsidiaries.*
 
-.. _`Security for software update in Clear Linux* OS`: https://clearlinux.org/blogs/security-software-update-clear-linux-os-intel-architecture
-.. _`Recent GNU* C library improvements`: https://clearlinux.org/blogs/recent-gnu-c-library-improvements
+.. _`Security for software update in Clear Linux* OS`: https://clearlinux.github.io/news-blogs/security-software-update-clear-linux-os.html
+.. _`Recent GNU* C library improvements`: https://clearlinux.github.io/news-blogs/recent-gnu-c-library-improvements.html
```
bryteise, line 144: Guessing news-blogs was a new creation.

source/guides/clear/swupd.rst
```diff
@@ -344,7 +344,7 @@ Related topics
 
 .. _swupd source documentation: https://github.com/clearlinux/swupd-client/blob/master/docs/swupd.1.rst
 
-.. _Kata Containers: https://clearlinux.org/downloads/containers
+.. _Kata Containers: https://clearlinux.github.io/downloads/containers
```
bryteise, line 347: I don't think this is supposed to be changed since it is going to the content part of the site. Anything under downloads probably needs to not change.

source/guides/maintenance/deploy-at-scale.rst
```diff
@@ -242,8 +242,8 @@ Continuously test its use; Automate its use by redeploying |CL| and
 application on new hosts. This naturally minimizes configuration drift,
 challenges your monitoring systems, and business continuity plans.
 
-.. _`Downloads`: https://clearlinux.org/downloads/
-.. _`Containers`: https://clearlinux.org/downloads/containers
+.. _`Downloads`: https://clearlinux.github.io/downloads/
+.. _`Containers`: https://clearlinux.github.io/downloads/containers
```
bryteise, line 246: I don't think this is supposed to be changed since it is going to the content part of the site. Anything under downloads probably needs to not change.

source/guides/maintenance/download-verify-decompress.rst
```diff
@@ -173,4 +173,4 @@ Image types
 .. include:: ../../reference/image-types.rst
    :start-after: incl-image-filename-end:
 
-.. _image: https://clearlinux.org/downloads
+.. _image: https://clearlinux.github.io/downloads
```
bryteise, line 176: I don't think this is supposed to be changed since it is going to the content part of the site. Anything under downloads probably needs to not change.

source/reference/image-types.rst
```diff
@@ -77,4 +77,4 @@ Table 2 lists the currently available images that are platform specific.
      - Virtual Machine Disk for VMware\* platforms including Player,
        Workstation, and ESXi.
 
-.. _images: https://clearlinux.org/downloads
+.. _images: https://clearlinux.github.io/downloads
```
bryteise, line 80: I don't think this is supposed to be changed since it is going to the content part of the site. Anything under downloads probably needs to not change.

source/reference/system-requirements.rst
```diff
@@ -32,7 +32,7 @@ persistent storage.
 For hardware with less resources, use the *live server* installer because it
 has a smaller memory footprint.
 
-See https://clearlinux.org/downloads for more download options.
+See https://clearlinux.github.io/downloads for more download options.
```
bryteise, line 35: I don't think this is supposed to be changed since it is going to the content part of the site. Anything under downloads probably needs to not change.

source/_themes/otc_tcs_sphinx_theme/cversions.html
bryteise: Ah you fixed this one.

bryteise:
One question and then a few comments about the downloads links and if they are changing.

source/tutorials/multi-boot/dual-boot-linux.rst
```diff
@@ -416,7 +416,7 @@ Ubuntu as our working example.
 
       sudo clr-boot-manager set-timeout 25
 
-.. _download the live desktop image: https://clearlinux.org/downloads
+.. _download the live desktop image: https://clearlinux.github.io/downloads
 
-.. _Downloads: https://clearlinux.org/downloads
+.. _Downloads: https://clearlinux.github.io/downloads
```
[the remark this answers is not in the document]
djklimes: if it starts with download.clearlinux it doesn't change.  If it starts with clearlinux.org it changes.  That is the correct link but it's still having issues because of the new cname that got put in clearlinux.github.io.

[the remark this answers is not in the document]
bryteise: Gotcha, makes sense.

bwarden:
Should we make these relative links where applicable so we're not embedding the domain name as much?

djklimes:
> Should we make these relative links where applicable so we're not embedding the domain name as much?

Part of the issues here is that while many of the "clearlinux" sites might seem like they are the same they aren't.  The https://www.clearlinux.org/clear-linux-documentation/ (this set of pages) is a different github set of pages then the original site (download.clearlinux.org, and community.clearlinux.org are two more sites).

bwarden:
> Part of the issues here is that while many of the "clearlinux" sites might seem like they are the same they aren't. The https://www.clearlinux.org/clear-linux-documentation/ (this set of pages) is a different github set of pages then the original site (download.clearlinux.org, and community.clearlinux.org are two more sites).

I get that, but this content is going under clearlinux.github.io, right? I see a lot of hardcoded https://clearlinux.github.io in there. But I do see that a lot of the actual self-referencing in the content is relative links. So maybe what's left is just stuff we want to be sure hard links back to our site even if it's an offline copy somewhere?

djklimes:
CNAME changes processed.  As long as the redirects are working this isn't needed.

PR #1274, opened 2024-05-26
Create change_kernel_boot.rst

sincorchetes:
Process to set a default kernel in your system.


pixelgeek (commented):
probably need to drop this into the guides/kernel section, and make sure there's a pointer to it.

sincorchetes:
okay, I moved to the sources/guides/kernel section

sincorchetes:
Do you think it works fine? I cannot see it on the https://www.clearlinux.org/clear-linux-documentation/guides/clear/compatible-kernels.html

PR #1276, opened 2024-10-23
Fix KVM instructions for zsh users

bashtoni:
The instructions for downloading the KVM image fail under zsh because it tries to expand the `*` that is outside of the quotes.

Quoting the entire filename makes the command work on both zsh and bash.


alexjch (approved):

alexjch:
@bashtoni Thanks for your contribution. I had to modify your PR to add some configuration changes. I'm closing your PR and linking your [fix](https://github.com/clearlinux/clear-linux-documentation/pull/1278) with my modification for reference.

